PR #4, opened 2021-07-08
Context Injector

DavidLarsKetch:


DavidLarsKetch:
Rejecting in favor of non-configurable approach.

PR #9, opened 2021-07-30
removes Props moving functionality to Vue

DavidLarsKetch:


DavidLarsKetch:
@profsmallpine here's my stab at dropping `Props` in favor of `Vue` structuring everything according to a Vue app's reqs. The trade off is this default "props" key when what's provided by `Data` is not a `map[string]interface{}`. Right now, everything in `second-child` and `college-try` is because of the old `buildVueResponse`, so ensuring that pattern is followed appears as a suggestion in the comment to `Vue`.

PR #10, opened 2021-09-02
improved logger

DavidLarsKetch:
Commented out code is a potential change to the API wherein the `logger.LogContext` is included in a log statement through a builder-style pattern as opposed to a parameter in the actual logging method being called.

### DIFF
--- a/logger/context.go
+++ b/logger/context.go
@@ -1,0 +1,93 @@
+package logger
+
+import (
+	"bytes"
+	"encoding"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+var (
+	_ encoding.TextMarshaler = LogContext{}
+)
+
+type LogUser interface {
+	GetID() uint
+	GetEmail() string
+}
+
+type LogContext struct {
+	Data    map[string]interface{}
+	Error   error
+	Request *http.Request
+	User    LogUser
+}
+
+// MarshalText converts LogContext into a JSON representation, eliminating zero-value fields.
+//
+// Values in LogContext.Data that cannot be represented in JSON will cause an error to be thrown.
+//
+// MarshalText implements encoding.TextMarshaler.
+func (lc LogContext) MarshalText() ([]byte, error) {
+	m := make(map[string]interface{})
+	if lc.Data != nil {
+		m["data"] = lc.Data
+	}
+
+	if lc.Error != nil {
+		m["error"] = lc.Error.Error()
+	}
+
+	if lc.Request != nil {
+		r := make(map[string]interface{})
+		r["method"] = lc.Request.Method
+		r["url"] = lc.Request.URL.String()
+		r["header"] = lc.Request.Header
+		if ct := lc.Request.Header.Get("Content-Type"); ct == "application/json" {
+			j := make(map[string]interface{})
+			b := new(bytes.Buffer)
+			tee := io.TeeReader(lc.Request.Body, b)
+			if err := json.NewDecoder(tee).Decode(&j); err == nil {
+				r["json"] = j
+				lc.Request.Body.Close()
+				lc.Request.Body = io.NopCloser(b)
+			}
+		}
+
+		if lc.Request.Form != nil {
+			r["form"] = lc.Request.Form
+		}
+
+		if len(r) > 0 {
+			m["request"] = r
+		}
+
+	}
+
+	if lc.User != nil {
+		u := make(map[string]interface{})
+		if id := lc.User.GetID(); id != 0 {
+			u["id"] = id
+		}
+		if email := lc.User.GetEmail(); email != "" {
+			u["email"] = email
+		}
+		if len(u) > 0 {
+			m["user"] = u
+		}
+	}
+
+	return json.Marshal(m)
+}
+
+// String stringifies LogContext as a JSON representation of it.
+func (lc LogContext) String() string {
+	b, err := json.Marshal(lc)
+	if err != nil {
+		fmt.Println(err)
+		return ""
+	}
+	return string(b)
+}
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,80 +1,173 @@
 package logger
 
-import "log"
+import (
+	"log"
+	"os"
+	"runtime"
+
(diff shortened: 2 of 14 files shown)


DavidLarsKetch:
Check out converting to this version of the logger here: https://github.com/xy-planning-network/college-try/pull/201/

logger/logger.go
```diff
@@ -1,80 +1,173 @@
...
+		skip: 2,
+		env:  getEnvOrString("ENVIRONEMNT", "DEVELOPMENT"),
+		l:    logger,
+		ll:   LogLevelInfo,
+	}
+	for _, opt := range opts {
+		opt(l)
+	}
+
+	if sentryDsn := os.Getenv("SENTRY_DSN"); sentryDsn != "" {
```
DavidLarsKetch, line 86: Here we swap out the default `TrailsLogger` for `SentryLogger` if the env var is set.

logger/logger.go
```diff
@@ -1,80 +1,173 @@
 package logger
 
-import "log"
+import (
+	"log"
+	"os"
+	"runtime"
+
+	"github.com/fatih/color"
```
DavidLarsKetch, line 8: Opted for this package to have cross-compiled color output.

logger/sentry.go
```diff
@@ -0,0 +1,122 @@
...
+	l Logger
+	//ctx LogContext
+}
+
+// NewSentryLogger constructs a SentryLogger based off the provided TrailsLogger.
+func NewSentryLogger(tl *TrailsLogger, dsn string) Logger {
+	err := sentry.Init(sentry.ClientOptions{
+		Dsn:          dsn,
+		Environment:  tl.env,
+		IgnoreErrors: []string{"write: broken pipe"},
```
profsmallpine, line 20: Make this configurable? I didn't want broken pipe errors on archive given how much networking I'm doing, but I don't think you'd want that elsewhere.

logger/logger.go
```diff
@@ -1,80 +1,173 @@
...
+	ll   LogLevel
 }
 
 // NewLogger constructs a TrailsLogger.
-func NewLogger() Logger { return &TrailsLogger{log.Default()} }
+//
+// Logs are printed to os.Stdout by default, using the std lib log pkg.
+// The default environment is DEVELOPMENT.
+// The default log level is DEBUG.
+func NewLogger(opts ...LoggerOptFn) Logger {
```
profsmallpine, line 74: Nitpick: but since you are going to be importing logger, you will call `logger.NewLogger`, which is redundant. I already have the "logger" info so I'm wanting to call `logger.New` here.

logger/context.go
```diff
@@ -0,0 +1,93 @@
...
+	Error   error
+	Request *http.Request
+	User    LogUser
+}
+
+// MarshalText converts LogContext into a JSON representation, eliminating zero-value fields.
+//
+// Values in LogContext.Data that cannot be represented in JSON will cause an error to be thrown.
+//
+// MarshalText implements encoding.TextMarshaler.
```
profsmallpine, line 32: Maybe you can help me understand why this is necessary?

logger/sentry.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	tl.skip = 3
+	return &SentryLogger{l: tl}
+}
+
+// Debug writes a debug log.
+func (sl *SentryLogger) Debug(msg string, ctx *LogContext) {
+	sl.l.Debug(msg, ctx)
+	//	sl.l.WithContext(sl.ctx).Debug(msg)
```
profsmallpine, line 35: Curious why you left this in?

logger/context.go
DavidLarsKetch: Why we need a marshaler or why I used `encoding.TextMarshaler`?

logger/sentry.go
DavidLarsKetch: For later reference should we want to update the API.

logger/context.go
profsmallpine: Yah, not having much context, just a noob question. Trying to understand why this was needed.

logger/sentry.go
profsmallpine: Maybe discuss the pros/cons and then move forward with an option now? These kind of maybes rarely get revisited and they add cruft when trying to understand what does exist

logger/context.go
DavidLarsKetch: Not a newb question at all - the primary reason is I wanted to include the `*http.Request` in the log statement, but that struct has functions that cannot be serialized into JSON. So, manually marshal it so I can skirt around those. Since I'm already doing that, I go ahead and hand-write the `map[string]interface{}` that is eventually passed to `json.Marshal` to be super explicit about the schema of this object.

profsmallpine:
@DavidLarsKetch Main issue I've noticed is not having the stack trace. I haven't looked into how much extra configuring this would be, but I think it's part of MVP

DavidLarsKetch:
@profsmallpine On sentry or stack traces on errors? I'm not seeing any issues with sentry at the moment.

profsmallpine:
@DavidLarsKetch Sorry, should have been more specific. If I have an error logged from the generic response, I don't know where it actually came from

logger/logger.go
```diff
@@ -1,80 +1,173 @@
...
+	return logger
+}
+*/
+
+// log executes printing the log message,
+// including any context if available.
+func (l *TrailsLogger) log(colorizer func(string, ...interface{}) string, level LogLevel, msg string, ctx *LogContext) {
+	// TODO(dlk): have skip be configurable or implement some logic
+	// like https://github.com/sirupsen/logrus/blob/b50299cfaaa1bca85be76c8984070e846c7abfd2/entry.go#L178-L213
+	_, file, line, _ := runtime.Caller(l.skip)
```
DavidLarsKetch, line 157: @profsmallpine I think your question about stack trace is caller file & line. Because a logger is instantiated app-wide, there's only one skip count we can use. But, given there are different depths of the call site in the stack, the sample code referenced here shows a dynamic approach that would be a good reference; essentially, get to the caller of the first package that is outside `trails`.

PR #15, opened 2022-01-25
Trails Ranger

DavidLarsKetch:
TODO:
- [x] ~~`embed` default FS~~
  - The `go:embed` directive looks for files to embed from `ranger/` or wherever the directive is. From the docs:
    > The patterns are interpreted relative to the package directory containing the source file.
- [x] ~~`HandlerGroup` struct?~~
  - Currently have similar structs receiving handler methods
  - There isn't a clear pattern here since a module importing a type from `ranger` would need its own local type to add methods onto
- [ ] `keyring`: throw it away? Better pattern for passing around shared keys?
  - Look into removing session key from `keyring.Keyringable` and instead implement that simplified interface in `session` which would include its own methods
- [ ] Simplify route setup even further?
- [x] Review exported/unexported fields on `Ranger`
- [x] Reassess `WithSessionStore`/`DefaultSessionStore` and use of `middleware.CurrentUser`
  - `middleware.CurrentUser` attempts to use `defaultUserStorer`, which can be configured incorrectly due to no DB being available or silently fails because the application user does not embed `trails.User`
  - sessions do not require the `currentUser` pattern, however, so eliding `middleware.CurrentUser` if a `UserStorer` is not available may be simplest
  - create Redis-backed or flat-file `UserStorer`?
  - refactor `WithSessionStore` to be about setting up user-based sessions, which entails included a `UserStorer` as well as `SessionStorer`


http/example/main.go
```diff
@@ -52,7 +52,7 @@ func itsHammerTime() int64 { return time.Now().UnixNano() }
 // Handler shares the initialized Responder across all example responses.
 type Handler struct {
 	*Responder
-	Ring ctx.KeyRingable
+	Ring keyring.Keyringable
```
profsmallpine, line 55: As far as I know, this is a naming convention we've run with (keyring/keyable). 1. I like it! 2. Are there any corollaries in django/rails land so that this language is a little more shared across web framworks.

ranger/default.go
```diff
@@ -0,0 +1,390 @@
...
+)
+
+const (
+	// Base URL defaults
+	baseURLEnvVar  = "BASE_URL"
+	defaultBaseURL = "http://localhost"
+
+	// Environment defaults
+	envFile   = ".env"
+	envEnvVar = "ENVIRONMENT"
```
profsmallpine, line 30: I get the envEnv, but I also think envVar is shorter and clearer

ranger/default.go
```diff
@@ -0,0 +1,390 @@
...
+	u, err := url.ParseRequestURI(envVarOrString(baseURLEnvVar, defaultBaseURL))
+	if err != nil {
+		return func(_ *Ranger) (optFollowup, error) {
+			return nil, fmt.Errorf("%w: failed parsing %s: %s", ErrBadConfig, baseURLEnvVar, err)
+		}
+	}
+
+	return func(rng *Ranger) (optFollowup, error) {
+		args := []template.ParserOptFn{
+			template.WithFn(template.Env(rng.Env.String())),
```
profsmallpine, line 204: If we end up supporting server rendered components, might be nice to include Masterminds/sprig here or something similar.

ranger/default.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+
+		for _, opt := range opts {
+			args = append(args, opt)
+		}
+
+		auth := os.Getenv(sessionAuthKeyEnvVar)
+		if auth == "" {
+			k := fmt.Sprintf("%x", securecookie.GenerateRandomKey(64))
+			if err := writeToEnvFile(sessionAuthKeyEnvVar, k); err != nil {
```
profsmallpine, line 339: I appreciate the ease of use here, but I think we ought to make this difficult for the user. I'd gander a public review would have all sorts of qualms with writing to the .env. We also want to support multiple keys for rotation purposes

ranger/default.go
```diff
@@ -0,0 +1,390 @@
...
+func defaultServer() *http.Server {
+	return &http.Server{
+		Addr:         ":3000", // TODO(dlk): configurable
+		ReadTimeout:  envVarOrDuration(serverReadTimeoutEnvVar, DefaultServerReadTimeout),
+		IdleTimeout:  envVarOrDuration(serverIdleTimeoutEnvVar, DefaultServerIdleTimeout),
+		WriteTimeout: envVarOrDuration(serverWriteTimeoutEnvVar, DefaultServerWriteTimeout),
+	}
+}
+
+// writeToEnvFile appends the key value pair to the default env var file, .env
```
profsmallpine, line 375: 😮

ranger/doc.go
```diff
@@ -0,0 +1,64 @@
+/*
+
+Package ranger initializes and manages a trails app with sane defaults and custom configuration.
+
+Ranger
+
+The main entrypoint to package ranger is the Ranger type.
+A Ranger ought to be constructed with NewRanger and any options required by your application.
+Calling NewRanger() - without any options - is valid and fully operational.
```
profsmallpine, line 9: 🥳

ranger/doc.go
```diff
@@ -0,0 +1,64 @@
...
+a resp.WithParser() functional option may be passed
+to resp.NewResponder or this package's DefaultParser.
+
+Environment Variables
+
+Constructing a new *Ranger with NewRanger reads from the OS's environment variables.
+With that, trails leverages a file called .env to pickup configuration.
+A file named .env ought to exist
+in the same directory as the compiled trails' app's executable.
+If it does not, package Ranger will attempt to create it and write default values to it.
```
profsmallpine, line 46: I think we ought to not write to the .env. Should always be a read-only file IMO

profsmallpine (commented):
Major upgrade to trails! 😓

PR #25, opened 2022-10-24
Adds auth interface and implementation

profsmallpine:
This PR consolidates handling of JWT tokens and fetching google user's from an oauth token. Adding SSO to archive duplicates the logic from tax so I'm looking to add it to trails. Let me know if this direction make sense or if y'all prefer some copy pasta instead. I'll add tests if this gets general approval.


auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+
+type AuthService interface {
+	AuthenticateJWT(v url.Values) (AppToken, bool, error)
+	FetchUser(token *oauth2.Token) (*goauth2.Userinfo, error)
+}
+
+// An AppToken is the structure to make JWT authentication requests.
+type AppToken struct {
+	sc              jwt.RegisteredClaims
+	AccountAddress  string `json:"firmAddress"`
```
seedifferently, line 20: these fields seem pretty implementation-specific to me. do they really belong in `trails`?

auth/auth.go
```diff
@@ -0,0 +1,47 @@
...
+	IsXypnMember    bool   `json:"isXypnMember"`
+	Email           string `json:"email"`
+	FirstName       string `json:"firstName"`
+	LastName        string `json:"lastName"`
+	Phone           string `json:"phone"`
+	PortalID        string `json:"portalId"`
+}
+
+// Valid asserts whether the AppToken is a valid JWT token.
+func (at *AppToken) Valid() error {
```
seedifferently, line 32: nit: suggest renaming to `IsValid`

PR #26, opened 2022-11-04
Responder renders default error template when Html errors

DavidLarsKetch:
## Problem statement
If `Html` encounters an error while attempting to render, nothing helpful happens for the user or developer. The browser may throw an error (like Chrome's "this page is broken") or simply show a blank page. The most likely error to occur is misuse of `resp.Fn`s passed to a `resp.Responder` method. Without a default error template, this can muddy the waters on avoiding bugs or pushing broken code. Instead, lets have a default error template.

## What this does
1. `*Responder.Html` renders a template (an "error template") provided by the user (through configuration) when `Html` hits an unhappy path.
2. `http/template.NewParser` merges many different `fs.FS` into a single cache of HTML templates that can be opened during runtime.
3. `ranger` provides a default error template by embedding it (and an `fs.FS`) at compile-time. This - and other future defaults - can be overwritten by matching the path to a template exactly. There are two different kind of filesystems that one could use to accomplish this: the OS and an embedded filesystem.

In non-Production environments, here's what developers see with that default error template:
Screenshot 2023-01-13 at 9 17 19 AM

Production looks like this, given `resp.Responder` is initialized with a `ContactErrMsg` (likely to be refactored to a value passed in a `ranger.Config`):
Screenshot 2023-01-13 at 12 12 43 PM

## Follow-ups
1. I'm using the URLs for logos & the favicon found on xyplanningnetwork.com. This is brittle, but there's no simple solution at the moment. `http/router` is constrained to OS-level files for assets. This may need to be reconsidered, or, opened up further so we can images, specifically, and maybe CSS files as well.
2. For non-production environments, let's see what additional stack trace information would help debugging.
3. Unit test conveniences. Check out `http/example/second-example/main_test.go` where it confirms the error template rendered. In our apps, web handler unit tests could check that the template rendered was _not_ the error template. Checking `500` does suffice at the moment, but may provide an extra guard.

TODO:
- [x] Make the default error template more purty
- [x] ~~Can the default template be approached more as a layout where the user provides a partial that gets shown to their end user?~~
- [x] ~~Include other default templates?~~

Some of the TODOs were passed on in an effort to narrow the scope of what's possible and instead provide the canonical one path developers can follow.


ranger/fs.go
```diff
@@ -0,0 +1,73 @@
+package ranger
```
DavidLarsKetch, line 1: This is where the magic is: a custom implementation of `fs.FS`. The files for `trails` are embedded whereas OS files are treated like a regular filesystem. You can have, essentially, multiple file trees with the same filepaths but different files served up. It prioritizes OS-level files (so the files in a trails project like `college-try`, etc.) over files made available in `trails` itself.

ranger/fs.go
```diff
@@ -0,0 +1,73 @@
...
+// The cache cannot become invalid at runtime since pkgDir is embedded.
+// If a file is removed from the OS during runtime,
+// then a reference to it from the cache returns the same error (fs.ErrNotExist)
+// as if the cache did not have that reference.
+func (vfs *virtualFS) Open(name string) (fs.File, error) {
+	// NOTE(dlk): while a concurrent routine could add a reference
+	// to the cache before this returns,
+	// let's err on the side of performance and not have this function
+	// blocking while waiting to read and only block when needing to write.
+	fn, ok := vfs.cache[name]
```
profsmallpine, line 42: For discussion: what about a cache only "filesystem" that uses similar logic and a slice of strings to load all templates into the cache at runtime?

ranger/default.go
```diff
@@ -251,7 +270,13 @@ func DefaultResponder(opts ...resp.ResponderOptFn) RangerOption {
...
 
 			if rng.p == nil {
-				if _, err := DefaultParser(os.DirFS("."))(rng); err != nil {
+				vfs := &virtualFS{
+					cache:  make(map[string]func(string) (fs.File, error)),
+					osDir:  os.DirFS("."),
+					pkgDir: pkgFS,
+					Mutex:  sync.Mutex{},
+				}
+				if _, err := DefaultParser(vfs)(rng); err != nil {
```
profsmallpine, line 279: Curious if this implementation of the vfs should live in http/template rather than being tried to ranger

profsmallpine (commented):
Getting this all tied together has felt so much more complex than the original `template.Execute..` logic, but seeing how you could pull together error pages with templates from trails is a nice direction for getting some wins out of the complexity.

ranger/default.go
DavidLarsKetch: That's a great question, I hadn't considered it due to the `go:embed` directive in `ranger/fs`. Again, the constraint with `embed` is wherever the embedded files live, the embed directive must be in the same directory as those files or a direct parent of them.  We see this with the embed directive existing in `ranger/` and the files in `ranger/tmpl/`. So, moving it to `http/template` would also mean moving the files there, for example `http/template/default`

This isn't a problem - just a question of where it makes more sense to have those templates live. Definitely not opposed to it, either. If we always use the `ranger` package to include defaults, I can see it getting messy. I'm thinking, for example, of the desire to move Sentry logging out of `logger`. Curious if these comments clarify the idea for you.

ranger/fs.go
DavidLarsKetch: Did it! 🎉

ranger/default.go
DavidLarsKetch: Okay, I played around with it being in `http/template` and concluded against moving it there in favor of keeping it in `ranger` as a default. The main reason was the functions and data the template needs are outside the scope of `http/template`.

Here's where my thinking is heading. `ranger` is the single dumping ground for defaults values, configuration, etc. rather than the packages it imports. This ways, there's a single layer someone can review when behavior doesn't match expectations instead of defaults living in different layers. Don't like the default error template? Look at `ranger`. Don't like the default environment? Look at `ranger`. And so forth. `ranger` functions as the "glue" or "spaghetti" code between the other `trails` packages so their respective scopes stay focused.

DavidLarsKetch:
@profsmallpine before I jumped over to session store stuff, wanted to get this out of my head. It's ready for another review.

ranger/default.go
```diff
@@ -211,20 +224,28 @@ func DefaultParser(files fs.FS, opts ...template.ParserOptFn) RangerOption {
...
-			template.WithFS(files),
 			template.WithFn(template.Env(rng.env)),
 			template.WithFn(template.Nonce()),
 			template.WithFn(template.RootUrl(rng.url)),
-			template.WithFn("packTag", template.TagPacker(rng.env, files)),
 			template.WithFn("isDevelopment", func() bool { return rng.env.IsDevelopment() }),
 			template.WithFn("isStaging", func() bool { return rng.env.IsStaging() }),
 			template.WithFn("isProduction", func() bool { return rng.env.IsProduction() }),
+			template.WithFn("packTag", template.TagPacker(rng.env, files)),
+			template.WithFn("metadata", func(key string) string {
```
profsmallpine, line 238: Make these alphabetical

http/example/second-example/tmpl/error.tmpl
```diff
@@ -0,0 +1,42 @@
...
+    <main>
+      <p>
+        Hey there, I replace the default error template provided by <code>trails/http/template</code>.
+      </p>
+      <p>
+      Because I match the filepath in <code>trails/http/template</code>,
+      that package prefers a user-provided file over ones the package provides.
+      </p>
+      <p>
+      Cool, huh?
```
profsmallpine, line 31: It is c00l!

http/resp/responder.go
```diff
@@ -254,7 +263,7 @@ func (doer *Responder) Json(w http.ResponseWriter, r *http.Request, opts ...Fn)
 	return nil
 }
 
-/* TODO(dlk): keep?
+/*
```
profsmallpine, line 266: Heads up: https://github.com/xy-planning-network/second-child/blob/9b2437dba2d4edbdf94a7bc5cfeca49a2d5d007d/http/api/api.go#L41

Couple places we need to not nest data

http/resp/responder.go
```diff
@@ -409,6 +418,49 @@ func (doer *Responder) do(w http.ResponseWriter, r *http.Request, opts ...Fn) (*
...
+
+	if doer.templates.err == "" {
+		err = fmt.Errorf(
+			"%w: no error template provided, encountered while handling: %s",
+			ErrBadConfig,
+			err,
+		)
+		return err
+	}
+	b := doer.pool.Get().(*bytes.Buffer)
```
profsmallpine, line 434: Not super relevant to this PR, but I feel like the pool adds complexity with rendering without really solving anything for our use cases. It would be one thing if we went to the pool because we were hitting some thresholds.

http/resp/responder.go
```diff
@@ -409,6 +418,49 @@ func (doer *Responder) do(w http.ResponseWriter, r *http.Request, opts ...Fn) (*
...
+
+	tmpl, nested := doer.parser.Parse(doer.templates.err)
+	if nested != nil {
+		err = fmt.Errorf("%w: %s", nested, err)
+		doer.logger.Error(err.Error(), nil)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return err
+	}
+
+	nested = tmpl.Execute(b, map[string]any{"Contact": doer.contactErrMsg, "Error": err})
```
profsmallpine, line 446: I guess this is a spot where contactErrMsg could be empty. This is where config would step in to help make it clear what is required for ranger to really hum

http/resp/responder_test.go
```diff
@@ -362,11 +372,81 @@ func TestResponderHtml(t *testing.T) {
...
+				require.Nil(t, err)
+				require.Equal(t, http.StatusInternalServerError, w.Code)
+			},
+		},
+		// NOTE(dlk): some sleight-of-hand here -
+		// resp.Authed() is not used since it will error first before
+		// reaching the checks in *Responder.Html;
+		// this test verifies someone setting the Authed template
+		// via another path is gracefully handled.
+		{
```
profsmallpine, line 415: Loving these extra tests

http/template/fs.go
```diff
@@ -0,0 +1,50 @@
+package template
+
+import (
+	"fmt"
+	"io/fs"
+)
+
+// A mergeFS maps filepaths to fs.Open functions,
```
profsmallpine, line 8: What are you using for line length of comments, these look like minis?

profsmallpine (approved):
Reiterating my conclusion from before.. this is a cool expose of what ranger gives us

http/template/fs.go
DavidLarsKetch: Just using the ["semantic linefeed"](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) encouraged by [Go docs](https://tip.golang.org/doc/comment)

http/resp/responder.go
DavidLarsKetch: Yeah definitely, would love to see "webmaster" contact and a default flash message in `ranger.Config` to ensure these values are available deeper down.

http/resp/responder.go
DavidLarsKetch: Awesome, haven't had the use case in `college-try` but use this on small projects (like where I was doing server-sent events). I'll followup with another PR, get some unit tests in.

http/template/fs.go
profsmallpine: This is great, so many times i've changed a line and then had to change the next 6 lines of comments

DavidLarsKetch: I think I overdo it at times,
chopping sentences up
too much.
But then,
I realize it is poetry.

PR #28, opened 2022-11-29
Remove rate limiter from default middleware stack in ranger

profsmallpine:
Removed rate limiter in default stack

### DIFF
--- a/ranger/default.go
+++ b/ranger/default.go
@@ -179,9 +179,7 @@ func DefaultLogger(opts ...logger.LoggerOptFn) RangerOption {
 	args := []logger.LoggerOptFn{
 		logger.WithLevel(logLvl),
 	}
-	for _, opt := range opts {
-		args = append(args, opt)
-	}
+	args = append(args, opts...)
 
 	return func(rng *Ranger) (OptFollowup, error) {
 		return WithLogger(logger.New(args...))(rng)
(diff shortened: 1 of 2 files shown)


ranger/default.go
```diff
@@ -359,7 +352,7 @@ func DefaultRouter() RangerOption {
 			r := router.NewRouter(rng.env.String())
 			r.OnEveryRequest(mws...)
 			r.HandleNotFound(http.HandlerFunc(func(wx http.ResponseWriter, rx *http.Request) {
-				if strings.Index(rx.Header.Get("Accept"), "text/html") >= 0 && rx.URL.Path != rng.url.Path {
+				if strings.Contains(rx.Header.Get("Accept"), "text/html") && rx.URL.Path != rng.url.Path {
```
DavidLarsKetch, line 355: This is different behavior and more permissive than `Index`. What's the reason for the change?

profsmallpine: I can bump it back, just linter change

profsmallpine: I'm not tracking what you mean. Can you provide an example where the code would do something different. Contains is:

```
Index(s, substr) >= 0
```

under the hood

DavidLarsKetch: Ah, yes, my mistake. I meant for the comparison to be `strings.Index(...) == 0`. Oh well, we shouldn't change the behavior now, so `Contains` is correct.

PR #29, opened 2023-01-10
Profsmallpine/env update

profsmallpine:


profsmallpine:
Well, this got a little wild. I grepped for .env and realized that some of the package were relying on a string env. I thought that supporting both felt a little janky. I updated everything except for the logger given that I don't know how it's being consumed out there in the wild and with an impending review by seth on that in general. I also moved env to the root so that other places could import it. I could have either created a new errors file here to support that move or move that file. I felt that having some trails errors made a lot of sense for errors.Is checking.

ranger/default.go
```diff
@@ -297,7 +296,7 @@ func DefaultRouter() RangerOption {
 			}
 
 			mws := make([]middleware.Adapter, 0)
-			if rng.env == Production {
+			if rng.env == trails.Production {
```
DavidLarsKetch, line 299: `rng.env.IsProduction()`

ranger/ranger.go
```diff
@@ -43,13 +44,22 @@ type Ranger struct {
 	users     middleware.UserStorer
 }
 
-// New constructs a Ranger from the provided options.
+// Config holds the data that is requried to init a new ranger. This also allow us to
```
DavidLarsKetch, line 47: `requried`

Also, as a trails user, the second sentence is unclear to me. It read more as explanation for a trails dev, maybe move to a comment inside `New` below?

ranger/ranger.go
```diff
@@ -43,13 +44,22 @@ type Ranger struct {
...
+// guarantee that certain attributes have been hyrdrated when running the options.
+type Config struct {
+	Env trails.Environment
```
DavidLarsKetch, line 50: Given this is exposed to users, lets start documenting these fields for users to have clarity about what it means since the functional options had comments explaining them.

DavidLarsKetch (commented):
Love it!

My one hangup is `trails.Config`. Are you able to give a brief description of the problem your approach:
```go
env := trails.EnvVarOrEnv("ENVIRONMENT", trails.Development)
cfg := ranger.Config{Env: env}
opts := []ranger.Fn{/* user defined opts */}

r, _ := ranger.New(cfg, opts...)
```
solves that the existing one does not:
```go
opts := ranger.Fn{/* user defined opts */}
r, _ := ranger.New(opts...)
// use r.Env for other references
```
I know we've discussed previously, but seeing it in front of me, it is hard to see the win. Do we want users to be able to define their own "ENVIRONMENT" env var? Do we want users to have that required argument brought to the forefront for explicitness? Is it unclear what overwriting `r.Env` means after calling `New`?

ranger/default.go
profsmallpine: 🤦

DavidLarsKetch:
Thanks! Helps the discussion with all the cards on the table. I'm unsure if my comments below necessitate any code changes at the moment. I'm aware that the solution I want isn't arbitrarily introduced in our projects, whereas the specific changes around `ranger.Config` are. In any case, I agree this is headed in the right direction, but not close enough such that it wouldn't be used.

The biggest challenge we're facing is the chicken/egg problem you're identifying. Specifically, handlers, procs and services may need certain values/dependencies that ranger can configure, while ranger may need certains procs or services (e.g., the database cxn) in order be intialized. Without resolving that, the "two approaches" vagueness Seth ran into persists, in my view.

Moving away from public consumption, the goal is to provide a single path, as you mention. Replacing functional options with configuration fields doesn't inherently solve that. For something like a "contact us" recipeent, that replacement _can_ and likely is the solution considering there's just 1 constant value always used and your app simply needs to define it:
```go
c := ranger.Config{ContactUs: "hello@xyplanningnetwork.com"}
```

But, for something like an environment, which has different values, I see the solution as restricting configuration to env vars. To use the environment example, this PR (and the pre-existing code) allows me to do:
```go
--- .env ---
MY_SPECIAL_ENVIRONMENT=DEVELOPMENT

--- main.go ---
import (
  "github.com/xy-planning-network/trails"
  "github.com/xy-planning-network/trails/ranger"
)
func main() {
  c := trails.EnvVarOrEnv("MY_SPECIAL_ENVIRONMENT", trails.Production)
  r, _ := ranger.New(c)
}
```
instead of
```go
--- .env ---
ENVIRONMENT=DEVELOPMENT

--- main.go ---
import "github.com/xy-planning-network/trails/ranger"
func main() {
  r, _ := ranger.New()
}
```
trying to imagine I were Seth, I would see the config struct as requiring a `trails.Environment` and grok that, but then be left to decide how I get one of those. Do I just do:
```go
e := trails.Environment("DEVELOPMENT") // maybe obviously no? maybe not?
```
or
```go
e := trails.EnvVarOrEnv("ENVIRONMENT", trails.Environment) // does it have to be ENVIRONMENT?
```
I'm being a bit dense here on purpose since its not always obvious which kind of thing will strike someone as unclear. An environment may not be that kind of thing, but a user session storer probably is.

In any case, the goal (likely to be achieved later) I'm proposing is:
```go
--- ranger/ranger.go ---
func New(opts ...RangerOption) (*Ranger, error) {
  r := new(Ranger)
  ...
  r.Env = trails.EnvVarOrEnv("ENVIRONMENT", trails.Development)
  ...
  return r
}
```
The user has no choice but to stuff set the value in the "ENVIRONMENT" env var (NB: unit tests would do so with os.Setenv). Then, handlers, procs and services can read the value for their own initialization after ranger completes its.

To me, the biggest TODO that gets us to my proposed goal is refactoring sessions and session stores. It may need to utilize generics. It may need a `trails.User`. Not decided yet myself on the appropriate solution.

profsmallpine:
To linger on this a bit, I'm curious if ranger comes first, and everything else follows. The services + procs cannot get away from needing the env. This appears to be fine from archive's perspective. If so, we could move on your proposal (to let env be defined in the environment rather than a config object) now and fix the downstream changes.

DavidLarsKetch:
Cool. The one hangup for `college-try` (and looks like on `second-child` too) is the func for pulling users into sessions comes from the procs. Need to refresh my memory on the questions there before knowing how complex it'll be. If I'm not mistaken, we can get the updates for environment in `trails` in the right order and merged in and I can follow up with user sessions later this week.

DavidLarsKetch (approved):
🚀

PR #30, opened 2023-01-12
Resolves redirect bug

DavidLarsKetch:
## Problem statement
Endpoints that use the `AuthorizeApplicator` were not redirecting properly due to a `401` status code returning and not a proper redirect.

## What this does
The existing solution was too brittle/narrow. Take a look at the unit tests which support both the header supplying many values and it not. The latter - that the "Accept" header is seen as a single string - is now the expected, "happy path" case.

I am unsure why this stopped working and don't have an interest in scratching the surface too deep. At first I thought it was a Firefox thing, but it isn't. Both Firefox & Chrome's "Accept" headers are parsed as a single long string by `net/http` instead of multiple values.


profsmallpine (approved):
Headers are for the 🐦

PR #31, opened 2023-01-19
ranger & session.UserStorer refactor

DavidLarsKetch:
## Problem statement
The setup to get to a `session.UserStorer` that works with an application is not meeting our goals. First, it requires a tango with the application's procedures being configured first before a `*ranger.Ranger` can be constructed. Second, there's more than 1 way to set it up.

To be able to store the `currentUser` in a `*http.Request.Context`, an application must define how to retrieve the `currentUser` from a datastore before `middleware.CurrentUser` can then stash it in the context. For example, from `college-try`:
```go
func main() {
        ...
        procs := procedures.New()
        rng, err := ranger.New(
            ...
            ranger.WithUserSessions(userstorer{procs.User.GetByID}),
        )
        ...
}

type userstorer struct {
	get func(id uint) (domain.User, error)
}
func (u userstorer) GetByID(id uint) (middleware.User, error) {
	user, err := u.get(id)
	if err != nil {
		return nil, err
	}

	return middleware.User(&user), nil
}
```
This is cleaned up in `second-child` with:
```go
func main() {
        ...
        procs := procedures.New()
        rng, err := ranger.New(
            ...
            ranger.WithUserSessions(procs.User),
        )
        ...
}
```
This only proves the point: two ways to do one thing _and_ requires procedures to be setup before `ranger`.

## What this does
This PR introduces a config struct that is instantiated with an application's implementation of `middleware.User`: `Config[User middleware.User]`. Having this config struct accept a type parameter enables exposing that type to `ranger` and, by extension, the implementation of `middleware.UserStorer` it sets up.

How `ranger` retrieves the `currentUser` from the datastore is now generic and gives 1 way to do it. It uses a custom implementation of `postgres.DatabaseService` that wraps `postgres.DatabaseServiceImpl` - save `FindByID` - and knows what concrete type to use to perform that retrieval via the generic `Config`.

`postgres/service.DatabaseServiceImpl` exports the underlying `*gorm.DB` so that `ranger` can use it for preloading all of the associations the application's user has. This avoids some sort of additional configuration to `Config` or whatnot. This may prove too brittle or too narrow. For `college-try` and `second-child`, this is sufficient. For `wall-to-wall`, [it is not](https://github.com/xy-planning-network/wall-to-wall/blob/f8b2c7cb7599846ea9a32f13a865ecf2fc0bd4aa/procedures/user.go#L54-L74). I will have to review the application to see if this is in fact necessary, but it very well may not be.

## Discussion
Check out its use in `college-try`: https://github.com/xy-planning-network/college-try/pull/524
- [x] Should `middleware.User` be the interface to satisfy for `currentUser`? Should this be a `ranger` defined interface, which embeds `middleware.User`, instead?
  - Added `RangerUser`
- Can we accept the single layer of preloading occurring or do we need this to be configurable?
- [x] What do we think of the alternate definition for `Config`?
  - Rejected!

### DIFF
--- a/ranger/config.go
+++ b/ranger/config.go
@@ -1,0 +1,50 @@
+package ranger
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/xy-planning-network/trails"
+	"github.com/xy-planning-network/trails/http/middleware"
+	"github.com/xy-planning-network/trails/postgres"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type Config[User RangerUser] struct {
+	// NOTE(dlk): Ranger can accept a type parameter also.
+	// Config was chosen to minimize proliferating generic type parameters
+	// in all Ranger methods or references to Ranger.
+	// Config ought to be restricted to New.
+}
+
+// defaultUserStore constructs a function matching the signature of middleware.UserStorer.
+// This function pulls the User from the db by ID,
+// preloading all top-level associations.
+func (Config[User]) defaultUserStore(db postgres.DatabaseService) middleware.UserStorer {
+	findByID := db.FindByID
+
+	// NOTE(dlk): if ranger.Ranger.db was a *postgres.DatabaseServiceImpl
+	// instead of *postgres.DatabaseService,
+	// the type assertion would not be necessary;
+	// we are not ready to commit to this inflexibilty, yet.
+	if db, ok := db.(*postgres.DatabaseServiceImpl); ok {
+		findByID = func(model, id any) error {
+			return db.DB.Preload(clause.Associations).First(model, id).Error
+		}
+	}
+
+	return func(id uint) (middleware.User, error) {
+		var user User
+		err := findByID(&user, id)
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			err = fmt.Errorf("%w: User %d", trails.ErrNotExist, id)
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		return user, nil
+	}
+}
--- a/ranger/config.go
+++ b/ranger/config.go
@@ -11,7 +11,7 @@ import (
 	"gorm.io/gorm/clause"
 )
 
-type Config[User RangerUser] struct {
+type Config[U RangerUser] struct {
 	// NOTE(dlk): Ranger can accept a type parameter also.
 	// Config was chosen to minimize proliferating generic type parameters
 	// in all Ranger methods or references to Ranger.
@@ -21,7 +21,7 @@ type Config[User RangerUser] struct {
 // defaultUserStore constructs a function matching the signature of middleware.UserStorer.
 // This function pulls the User from the db by ID,
 // preloading all top-level associations.
-func (Config[User]) defaultUserStore(db postgres.DatabaseService) middleware.UserStorer {
+func (Config[U]) defaultUserStore(db postgres.DatabaseService) middleware.UserStorer {
 	findByID := db.FindByID
 
 	// NOTE(dlk): if ranger.Ranger.db was a *postgres.DatabaseServiceImpl
@@ -35,7 +35,7 @@ func (Config[User]) defaultUserStore(db postgres.DatabaseService) middleware.Use
 	}
 
 	return func(id uint) (middleware.User, error) {
-		var user User
+		var user U
 		err := findByID(&user, id)
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			err = fmt.Errorf("%w: User %d", trails.ErrNotExist, id)
--- a/ranger/default.go
+++ b/ranger/default.go
@@ -3,8 +3,6 @@ package ranger
 import (
 	"context"
 	"embed"
-	"errors"
-	"fmt"
 	"io/fs"
 	"net"
 	"net/http"
@@ -22,7 +20,6 @@ import (
 	"github.com/xy-planning-network/trails/http/template"
 	"github.com/xy-planning-network/trails/logger"
 	"github.com/xy-planning-network/trails/postgres"
-	"gorm.io/gorm"
 )
 
 const (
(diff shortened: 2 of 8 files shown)


ranger/ranger.go
```diff
@@ -44,10 +47,49 @@ type Ranger struct {
 	users     middleware.UserStorer
 }
 
+/*
+type Config struct {
+	UserStore func(db postgres.DatabaseService) middleware.UserStorer
+}
+*/
```
DavidLarsKetch, line 54: This was another option. Instead of the `Config` itself having the type parameter, below, the commented out `func UserStorer` does. I opted for `Config` itself since there's no chance for misuse: you can't init a `Config` with instantiating it with a type. With this alternative, one could always leave `UserStore` `nil` unintentionally.

ranger/default.go
```diff
@@ -341,35 +338,14 @@ func DefaultRouter() RangerOption {
 			mws = append(
 				mws,
 				middleware.InjectSession(rng.sessions, rng.kr.SessionKey()),
+				middleware.CurrentUser(
+					rng.Responder,
+					rng.users,
```
profsmallpine, line 343: This naming has always thrown me off. I assume it's an array of users or something. Maybe `userFetcher` or something more explicit will improve the readability here.

profsmallpine (approved):
I'm curious where the "internalDB" comes from when setting up the procedures. It might be worth exposing `EmitDB` vs `EmitDBService` or maybe changing the database service interface to have a method that returns the internal db.. as well as `EmitEnvironment()` so that all dependencies are available for services/procedures (all eggs are available to the chickens?) at this stage of changes.

ranger/ranger.go
profsmallpine: I'm unsure here. The win is clear and an improvement. However, I've leaned towards generics being a loss for golang. Only use it when you can't solve a problem with a more verbose option. This would lean towards a config struct with a nil guard in new. As always, happy to discuss further

profsmallpine:
Caveat here: this was a rushed review. In retrospect, I need to review generics a bit and pull this code down to play around a bit. I'll do that next week so take this feedback with a grain of🧂

DavidLarsKetch (commented):
@profsmallpine refactored a bit more. I was able to eliminate `internalDB` to clear up any confusion there - there's a more surgical approach.

With regards to generics, it's a tough call. My approach is to keep it constrained to `Config` and its methods and the `*ranger.Ranger` constructor `New`. This adds verbosity with tracking values throughout setup, but I don't want to see `Ranger` have to accept a type parameter.

Throwing together a followup PR (#32) that shows where `Config` is headed which may clarify the end goal.

ranger/config.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	"github.com/xy-planning-network/trails"
+	"github.com/xy-planning-network/trails/http/middleware"
+	"github.com/xy-planning-network/trails/postgres"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type Config[User RangerUser] struct {
```
profsmallpine, line 14: Nitpick: for some reason, my mind gets misdirected by the type name of `User`. I prefer the style of a single letter `config[U RangerUser]` and the cascading changes.

ranger/config.go
```diff
@@ -0,0 +1,50 @@
...
+	// NOTE(dlk): Ranger can accept a type parameter also.
+	// Config was chosen to minimize proliferating generic type parameters
+	// in all Ranger methods or references to Ranger.
+	// Config ought to be restricted to New.
+}
+
+// defaultUserStore constructs a function matching the signature of middleware.UserStorer.
+// This function pulls the User from the db by ID,
+// preloading all top-level associations.
+func (Config[User]) defaultUserStore(db postgres.DatabaseService) middleware.UserStorer {
```
profsmallpine, line 24: This is a brilliant + simple way to get the fetching of the user into trails. Cool application of generics here 🥳

ranger/ranger.go
```diff
@@ -41,13 +47,13 @@ type Ranger struct {
 	shutdowns []ShutdownFn
 	srv       *http.Server
 	url       *url.URL
-	users     middleware.UserStorer
+	userstore middleware.UserStorer
```
profsmallpine, line 50: Gracias!

profsmallpine (approved):
I like where this is headed. w00t

profsmallpine:
In response to the differences w/wall-to-wall, I'm curious if the helper func should contain any extra complexity rather than trying to squeeze it into trails. We already need a helper to pull the user out of the session as the domain type. Might make sense (though a bit less efficient from a DB perspective) to do some extra DB activity in that same spot. It removes some of the magic here, and leaves the code that is setting up the user for every single request in the application code which feels a bit more explicit to me if I were to be debugging an app issue.

DavidLarsKetch:
> if I were to be debugging an app issue

This is definitely one of the biggest risks given GORM is actually performing the DB call. It could become extremely unclear if there's just some issue with how GORM's being used (e.g., its not picking up an association because of misconfig) or if `Ranger` is at fault (e.g., there's a legitimate bug in the package). If we keep the implementation of the DB call

---

Now, to counter my own hesitation, I do want to highlight some of the possible end results with having an app's currentUser type available. Your comment below touches on it:

> We already need a helper to pull the user out of the session as the domain type.

If we put the app's type for currentUser on `Ranger` itself (and not just `Config`), it can have a method like:
```go
func (r *Ranger[U RangerUser]) CurrentUser(ctx context.Context) (U, error) {
    u, ok := ctx.Value(trails.CurrentUserKey).(U)
    if !ok {
        return nil, domain.ErrNoUser
    }
    
    return u, nil
}
```
to render those helpers unnecessary.

Another refactor available is with `middleware.AuthorizeApplicator`. An app wouldn't need to construct its own as the type can be passed from `Ranger` to `middleware.NewAuthorizeApplicator` and then that factory (for lack of a better word) is exposed as a field on `Ranger`.

There's also the possibility of `resp` and its references to currentUser being better guarded and ergonomic if it knows what type is required.

Food for thought!

PR #32, opened 2023-01-20
ranger: config-based construction

DavidLarsKetch:
## Problem Statement
trails is not for public consumption but is only for internal use, now. Many components of the library were built with external users in mind. `ranger` was built to provide sane defaults, but also needed to make each and every one configurable.

### ranger
This PR simplifies the configurability of a `*ranger.Ranger` by forcing configuration through defaults, environment variables, and `Config`. There ought to be "one way" to setup a component of `*ranger.Ranger`. This PR removes multiple ways of setting up a component in favor of one path. The paths available now are (1) defaults set by ranger with env vars overwriting some values, (2) env vars setting required values, or (3) passing values through `Config`.

For now, `Config[U RangerUser]` leverages generics to pull a user from the database _as_ the type of user the application defines it as. This sprinkling of generics removes the need for an application to define a helper function for retrieving user from the database.

### toolbox
This is unrelated to everything else and was approved in #36. Given the amount of changes in flight, it felt prudent to branch off this feature branch instead of risk potential merge conflicts.

### Replacements
**http/keyring -> keys.go**
`keyring` was too abstract given we've landed on a circumscribed set of values passed around in `context.Context`. There's no lingering use case where `trails` needs a key defined by an application, so the package was removed. `http/middleware` and `http/resp` used `http/keyring` fairly extensively, but now need only reference to the specific keys from the trails root.

### Removals
**ranger/opt.go**
There are two remaining optional values application code can define and pass to `ranger`: migrations and shutdown functions. These are now captured in `Config.Migrations` and `Config.Shutdowns`, respectively. These two do not warrant an abstraction, so this PR removes the pattern entirely.

**http/example**
It isn't trivial to maintain these with `ranger` expecting to connect to a DB. `college-try`, `second-child` and so forth with be our examples.

**http/middleware/responder.go**
A middleware I used when using trails for personal projects, but not needed for ours 😢 

## What this doesn't do
It is worth noting that the responsibility of constructing defaults still resides in `ranger`. Here's a simple example. `http/resp` is the only package that cares about `http/template`. All `ranger` does is pass along the latter to the former. So, why doesn't `NewResponder` simply construct the `template.Parser` it needs?

The main reason to maintain this design decision is to force as much default configuration into one place. If there's some unexpected behavior by a component, one need only look into `ranger` to find it. Otherwise, a patchwork may result where `ranger` looks at _these_ env vars whereas `http/resp` looks at _those_ and the whole becomes a confused mess.

There are exceptions to the above and the intention is to weed all of those out.


ranger/default.go
```diff
@@ -27,10 +28,12 @@ const (
 	baseURLEnvVar = "BASE_URL"
 
 	// App metadata
-	appDescEnvVar   = "APP_DESCRIPTION"
-	defaultAppDesc  = "A trails app"
-	appTitleEnvVar  = "APP_TITLE"
-	defaultAppTitle = "trails"
+	appDescEnvVar    = "APP_DESCRIPTION"
+	defaultAppDesc   = "A trails app"
```
DavidLarsKetch, line 32: These "app metadata" env vars are good examples of required ones that do not have a useful default value. There's a guard that needs to be put in place to validate the values pulled from the environment are useful. Undecided where that should live at the moment.

You can contrast this with bad values for DB connection env vars just means the connection fails and the user gets the appropriate error message.

DavidLarsKetch:
@profsmallpine if you want to review, this PR contains all the previous branches with the addition of [this penultimate commit](https://github.com/xy-planning-network/trails/pull/32/commits/079f907eb2acc45ccb5d1ed51950ea7f2a945daf) putting a nail in the coffin. Barring overlooking anything, this finishes refactoring out all the flexible ways ranger was configurable in favor of the "one way" philosophy.

DavidLarsKetch:
@seedifferently if you're curious to check out the major refactor here, the pertinent changes are in `ranger/`, specifically `ranger/ranger.go` and the replace of `ranger/opt.go` with `ranger/config.go`. Probably easiest to chat any specific changes, but the overall gist is to (1) eliminate being able to overwrite defaults and (2) have ranger be configured "first" and then expose its components to an application. 2 will make more sense when I get a branch up showing how these changes work in `college-try`.

http/middleware/current_user.go
```diff
@@ -29,14 +29,14 @@ type UserStorer func(id uint) (User, error)
 // CurrentUser checks whether the "Accept" MIME type is "application/json"
```
seedifferently, line 29: > CurrentUser checks whether the "Accept" MIME type is "application/json" and write a status code if so.

crusty?

http/middleware/current_user.go
```diff
@@ -119,7 +119,7 @@ func RequireUnauthed(key keyring.Keyable) Adapter {
 	}
 }
 
-/// RequireAuthed returns a middleware.Adapter that checks whether a User is authenticated,
+// / RequireAuthed returns a middleware.Adapter that checks whether a User is authenticated,
```
seedifferently, line 122: 🧹

http/middleware/current_user.go
DavidLarsKetch: I think more just imprecise. `handleErr` below is where this toggling occurs.

ranger/default.go
```diff
@@ -97,331 +95,151 @@ var (
...
+func defaultRouter(
+	env trails.Environment,
+	baseURL *url.URL,
+	responder *resp.Responder,
+	mws []middleware.Adapter,
+) router.Router {
+	route := router.NewRouter(env.String())
+	route.OnEveryRequest(mws...)
+	route.HandleNotFound(http.HandlerFunc(func(wx http.ResponseWriter, rx *http.Request) {
+		if strings.Contains(rx.Header.Get("Accept"), "text/html") && rx.URL.Path != baseURL.Path {
```
seedifferently, line 203: Curious why a redirect was chosen as the default behavior for a `NotFound` (when accepting html) vs the typical `404` response?

ranger/ranger.go
```diff
@@ -41,69 +37,82 @@ type Ranger struct {
...
-		if fn != nil {
-			followups = append(followups, fn)
-		}
+	r.metadata, err = newMetadata()
+	if err != nil {
+		return nil, err
 	}
 
-	r.userstore = cfg.defaultUserStore(r.db)
+	url := trails.EnvVarOrURL(baseURLEnvVar, defaultBaseURL)
```
seedifferently, line 71: nit: I'd prefer something like `rootURL` here (and in `defaultResponder`) to be more explicit and to avoid name confusion/collision with `net/url`.

toolbox.go
```diff
@@ -0,0 +1,44 @@
+package trails
```
seedifferently, line 1: I'm curious what the rationale was for this sort of thing needing to live in `trails` vs the parent app?

DavidLarsKetch: The intention is to use it in `second-child`.

http/session/error.go
```diff
@@ -3,7 +3,6 @@ package session
 import "errors"
 
 var (
-	ErrFailedConfig = errors.New("failed config")
-	ErrNotValid     = errors.New("not valid")
-	ErrNoUser       = errors.New("no user")
+	ErrNotValid = errors.New("not valid")
```
profsmallpine, line 6: Put these in the root rather than having errors defined per package?

keys.go
```diff
@@ -0,0 +1,63 @@
...
+type ByKey []Key
+
+var _ sort.Interface = ByKey([]Key{})
+
+func (k ByKey) Len() int           { return len(k) }
+func (k ByKey) Swap(i, j int)      { k[i], k[j] = k[j], k[i] }
+func (k ByKey) Less(i, j int) bool { return string(k[i]) < string(k[j]) }
+
+// UniqueSort sorts, uniques and removes zero keys.
+func (k ByKey) UniqueSort() ByKey {
```
profsmallpine, line 38: It looks like this is used in `WithCtxKeys`. I'm curious how this came to be? Is it trying to make sure this is as tidy as possible? Or is it preventing an issue you bumped into? Same with `type ByKey []Key` in general

toolbox_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/xy-planning-network/trails"
+)
+
+func TestToolboxFilter(t *testing.T) {
+	// TODO
```
profsmallpine, line 11: This need to be revisited before merging?

profsmallpine (commented):
Took another glance before our conversation later today

http/session/error.go
DavidLarsKetch: I'm a bit undecided on the utility of having the root repo providing all sentinel errors for every other package. In line with that, the introduction of multiple error wrapping in Go 1.20 gives us opportunity to reflect on the strategy for errors in trails.

PR #34, opened 2023-01-31
Keyring & session refactor

DavidLarsKetch:
## Problem statement
There were two highly flexible components to the `trails` library that have not proven useful: `keyring` and the different interfaces defining the behavior of a `session.Session`.

### `keyring`
The `keyring` package was designed to enable exposing an apps own context keys to `trails` or from one package in an app to another package in an app without those keys always being exported. Interfaces defined the behavior and the package provided a default implementation `keyring.Keyring`.

Setting up a `Keyringable` meant having two required keys: a current user key and a session key. Each of these would get/set values from an `*http.Request.Context`.

### `session.Session`
`session.Sessionable`, `session.UserSessionable`, and `session.FlashSessionable` each described the behavior of the different ways sessions function. They were not dependent on one another. These were composed together in the `session.TrailsSessionable` interface.

### `keys.go`
This PR removes `http/keyring` in favor of `keys.go`

4 keys used throughout our projects are now exported by the `trails` root:
- `CurrentUserKey`: stashes the currentUser
- `IpAddrKey`: stashes the IP address of a request
- `RequestIDKey`: stashes a UUID for each unique request
- `SessionKey`: stashes the session

These keys are `const`s so there is no danger of an application overwriting their value.

Oblique references to `keyring.Keyable` are replaced with the appropriate key from `keys.go`. This is notable in `http/session` & `http/middleware`. Various functions and constructors required an implementation of the interface whereas now, specific keys from the above 4 are directly referenced.

### `session` refactor
This PR removes all of the session interfaces in favor a single struct: `session.Session`.

Previously, unit tests could implement whatever session interface necessary to make their unit tests work. Now, this PR makes a stub implementation of `SessionStorer` available that outputs `session.Session`, but a strange stub of one. What it does do was driven by the needs of unit tests in `trails`. It remains to be seen whether this will be a prudent strategy as it'll be incumbent of `trails` devs to know that stub can produce fairly exotic errors.

### `http/example` deleted
`http/example` was designed for public consumption and not our own internal uses. We have example uses of `trails` in our private repos. With `ranger` requiring the fuller suite of services, these toy examples would become too complicated and provide no additional illumination of `trails`'s features not already found in our private repos. So, we say goodbye.

## TODO
- [x] update `http/example` uses


ranger/default.go
```diff
@@ -191,8 +194,10 @@ func defaultResponder(l logger.Logger, url *url.URL, p template.Parser, kr keyri
 		resp.WithErrTemplate(defaultErrTmpl),
 		resp.WithUnauthTemplate(defaultUnauthedTmpl),
 		resp.WithVueTemplate(defaultVueTmpl),
-		resp.WithSessionKey(kr.SessionKey()),
-		resp.WithUserSessionKey(kr.CurrentUserKey()),
+	}
+
+	if len(ctxKeys) > 0 {
```
profsmallpine, line 199: I had imagined this would get kaput as well since we didn't end up using anything outside request/ip/session/current-user?

profsmallpine (approved):
Overall, this direction lgtm, lots of questions around "should this exist anymore" but that might be better in a live review or in some sort of final review where we get more nit-picky. Lots of things changing though so a little worried about the review process and losing the signal as things pile on top of each other.

ranger/default.go
DavidLarsKetch: Yeah the use case is in `second-child` and `wall-to-wall` which inject data into `*http.Request.Context` in a middleware to be included when `Vue` is is called for an `Html`.

DavidLarsKetch:
Yeah it's a lot in the air. I guess I'd ask whether you're reviewing `trails` or the changes in the PR. If you want to review `trails`, it seems good to wait for the other major changes to land. For example, whether `middleware.InjectResponder` should exist can happen then. Alternatively, if you don't like, for example, how I construct a name for sessions off the app title, let's settle that now. Maybe that's too narrow of a suggestion? Your call!

ranger/default.go
profsmallpine: I believe that got removed in the trails refactor from archive so just wall-to-wall which looks like it could just be a middleware and not need the context.

DavidLarsKetch: Not sure I follow on the suggestion. For `wall-to-wall`, a [middleware injects values](https://github.com/xy-planning-network/wall-to-wall/blob/f8b2c7cb7599846ea9a32f13a865ecf2fc0bd4aa/http/middleware.go#L12-L24) which [`resp.Vue` pulls](https://github.com/xy-planning-network/trails/blob/bd81c167795e335e3149741bf03f7a591ec6dfe2/http/resp/response.go#L350-L355).

DavidLarsKetch: Now that I'm thinking about it, a simplification would be to add another key to `keys.go` that is used for this special use case and get rid of `CtxKeys`.

PR #35, opened 2023-02-03
Replaces Responder.ctxKeys with trails.AppPropsKey

DavidLarsKetch:
This simplifies stashing additional values in a `*http.Request.Context` by exposing a single key to use for performing that getting & setting: `AppPropsKey`.

Also removes `http/middleware.InjectResponder` in favor of "one way to do things" philosophy 😢


profsmallpine (approved):
Nice more generic ;) solution for passing data to vue 🎆

PR #37, opened 2023-02-27
Update PG connections to fallback to default SSL mode

seedifferently:
Connections to Postgres should fallback to using the default `sslmode` param instead of `disabled`.

### DIFF
--- a/postgres/postgres.go
+++ b/postgres/postgres.go
@@ -9,17 +9,19 @@ import (
 	"gorm.io/gorm/schema"
 )
 
-const cxnStr = "host=%s port=%s user=%s dbname=%s sslmode=disable password=%s"
+// PG Docs: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS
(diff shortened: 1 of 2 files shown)


DavidLarsKetch (commented):
Do you see any value in this being configurable from an env var like the other ones are?

seedifferently:
@DavidLarsKetch yeah makes sense to keep it consistent, thanks

DavidLarsKetch (approved):

profsmallpine:
1. 🥂 for your first trails contribution @seedifferently 
2. I am using URL on the struct to change archive. I'd say let's close this PR and possibly block string building outside of dev once we know more from archive devops works in the coming months

seedifferently:
but can you actually call it a `contribution` if it's closed instead of merged?

DavidLarsKetch:
Eh... I lean towards merge. It makes me squeamish now that I realize the default is `sslmode=disable` instead of PostgreSQL's default `sslmode=prefer`. Just seems like a footgun to not forget about.

PR #38, opened 2023-03-31
Ranger loads default layout templates

DavidLarsKetch:
## Problem statement
`ranger` provides 2 [`fs.FS`](https://pkg.go.dev/io/fs#FS) to `http/template` in order to parse its own embedded files and those provided by the application (with the latter overriding any default templates in ranger). The initial use case was to ensure an error page is made as a backstop for unruly `*http/resp.Responder.Html` calls. We can push this paradigm further by including the layout templates every project uses.

## What this does
This PR creates a new `ranger/tmpl/layout` dir with 5 files, more on those files below. By doing this, development can begin immediately without needing to copy these files from another repo. Only when some additional markup diverging from the minimal defaults is needed would someone begin developing their own layout templates.
- `unauthenticated_base.tmpl` which renders in `http/resp.Unauthed` calls
- `authenticated_base.tmpl` which renders in `http/resp.Authed` calls
- `additional_scripts.tmpl` which injects `additionalHeadScripts` and `additionalBodyScripts` into the above 2 templates
- `vue.tmpl` which renders in `http/resp.Vue` calls
- `vue_scripts.tmpl` which injects `vueScripts` into the above template

For `college-try`, `second-child` & `wall-to-wall`, `additional_scripts.tmpl` would be need to be created in order to throw in all the 3rd party scripts those applications have. In `college-try`'s case, it needs `vueScripts` to throw in a bit more once the Vue app completes loading.

NB: it's important to recall that overriding any or all of the above files means (1) placing those files in a directory called `tmpl/layout`, (2) embedding the directory in your application, and (3) exposing that embedded `fs.FS` in app startup in `ranger.Config`.

### DIFF
--- a/http/resp/responder.go
+++ b/http/resp/responder.go
@@ -52,6 +52,8 @@ type Responder struct {
 	rootUrl *url.URL
 
 	templates struct {
+		additionalScripts string
+
 		// Root template to render when user is authenticated
 		authed string
 
@@ -64,6 +66,8 @@ type Responder struct {
 
 		// Vue template to render when rendering a Vue app
 		vue string
+
+		vueScripts string
 	}
 }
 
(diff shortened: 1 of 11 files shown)


profsmallpine (approved):
🤌

PR #39, opened 2023-04-21
replaces logger.Logger in middleware.LogRequest with slog.Logger

DavidLarsKetch:
## Problem statement
There are 2 problems this PR moves the needle towards solving.

1. Including more data in HTTP router handling logs. Doubling down on AWS, we want to know more about how HTTP requests are handled in our application and be able to trace those requests more reliably. As of now, our implementation of this functionality in the `LogRequest` middleware provides limited information.

2. Our logging lacks ergonomics. With a [`slog`](https://pkg.go.dev/golang.org/x/exp/slog) accepted for implementation in the std lib, let's jump on board.

## What this does
This PR makes 3 major changes:
1. It moves `LogRequest` to a "post-response" middleware. Some of the data we want to record can only be known after the request is handled (e.g., response status code), so must log the message after the application completes handling the request.

2. It includes data from the `*http.Request` and `http.ResponseWriter` as requested in [this Notion ticket](https://www.notion.so/xylabs/reLogging-refactor-984c89157f5f4458b06c5c08eafcd5de?pvs=4). This fills out the picture of what the request/response lifecycle looked like.

3. It uses `slog` in place of `trails/logger.Logger` for writing messages. This wasn't strictly necessary for problem 1 above, but this looked like a good, small use case showing some of the breadth of functionality in `slog`. Originally, I knocked out 1 by tab-delimiting the fields in the message, but the amount of new data made the message itself ungainly when compared to key-value based messages.
    - ⚠️ We ought to expect the API for `slog`to change before final acceptance into the std lib. We need not keep up with every iota, though. An incoming proposal PR will show how we'll take a two step process replacing `trails/logger` with `slog` where that first step will limit direct exposure to the `slog` API in our applications. ⚠️

### DIFF
--- a/http/middleware/get_ip_address.go
+++ b/http/middleware/get_ip_address.go
@@ -10,8 +10,6 @@ import (
 	"github.com/xy-planning-network/trails"
 )
 
-const IpAddrCtxKey = "trails/middleware/ip-address" // TODO(dlk): change to key provided by app?
-
 // An ipRange is a range of IP addresses.
 type ipRange struct {
 	start net.IP
--- a/http/middleware/log_request.go
+++ b/http/middleware/log_request.go
@@ -66,7 +66,7 @@ func LogRequest(handler slog.Handler) Adapter {
 
 // A LogRequestRecord represents the fields that a LogRequest
 type LogRequestRecord struct {
-	BodySize       int    `json:"responseBodySize"`
+	BodySize       int    `json:"bodySize"`
 	Duration       int64  `json:"duration"`
 	Host           string `json:"host"`
 	ID             string `json:"id"`
@@ -75,8 +75,8 @@ type LogRequestRecord struct {
 	Path           string `json:"path"`
 	Protocol       string `json:"protocol"`
 	Referrer       string `json:"referrer"`
-	ReqContentLen  int    `json:"requestContentLength"`
-	ReqContentType string `json:"requestContentType"`
+	ReqContentLen  int    `json:"contentLength"`
+	ReqContentType string `json:"contentType"`
 	Scheme         string `json:"scheme"`
 	Status         int    `json:"status"`
 	URI            string `json:"uri"`
@@ -98,7 +98,7 @@ func newRecord(w *requestLogger, r *http.Request) LogRequestRecord {
 	ip, _ := r.Context().Value(trails.IpAddrKey).(string)
 
 	return LogRequestRecord{
-		BodySize:       w.bodysize,
+		BodySize:       w.bodySize,
 		Host:           r.Host,
 		ID:             id,
 		IPAddr:         ip,
@@ -125,9 +125,9 @@ func (r LogRequestRecord) attrs() []slog.Attr {
 		slog.String("path", r.Path),
 		slog.String("protocol", r.Protocol),
 		slog.String("referrer", r.Referrer),
-		slog.Int("requestContentLength", r.ReqContentLen),
-		slog.String("requestContentType", r.ReqContentType),
-		slog.Int("responseBodySize", r.BodySize),
+		slog.Int("contentLength", r.ReqContentLen),
+		slog.String("contentType", r.ReqContentType),
+		slog.Int("bodySize", r.BodySize),
 		slog.String("scheme", r.Scheme),
 		slog.Int("status", r.Status),
 		slog.String("uri", r.URI),
@@ -138,7 +138,7 @@ func (r LogRequestRecord) attrs() []slog.Attr {
 type requestLogger struct {
 	http.ResponseWriter
 	status   int
-	bodysize int
+	bodySize int
 }
 
 func (rl *requestLogger) Header() http.Header { return rl.ResponseWriter.Header() }
@@ -149,7 +149,7 @@ func (rl *requestLogger) Header() http.Header { return rl.ResponseWriter.Header(
 func (rl *requestLogger) Unwrap() http.ResponseWriter { return rl.ResponseWriter }
 func (rl *requestLogger) Write(b []byte) (int, error) {
 	size, err := rl.ResponseWriter.Write(b)
-	rl.bodysize += size
+	rl.bodySize += size
 
 	return size, err
 
--- a/http/middleware/log_request.go
+++ b/http/middleware/log_request.go
@@ -2,44 +2,167 @@ package middleware
 
 import (
 	"net/http"
-	"strings"
+	"net/url"
+	"strconv"
+	"time"
 
-	"github.com/xy-planning-network/trails/logger"
+	"github.com/xy-planning-network/trails"
+	"golang.org/x/exp/slog"
 )
 
-// LogRequest logs the request's method, requested URL, and originating IP address
-// using the enclosed implementation of logger.Logger.
+const (
+	passwordParam     = "password"
+	contentLenHeader  = "Content-Length"
+	contentTypeHeader = "Content-Type"
+	referrerHeader    = "Referrer"
+	userAgentHeader   = "User-Agent"
+)
+
+// LogRequest logs the a LogRequestRecord using the provided handler.
 //
-// LogRequest scrubs the values for the following keys:
+// For the LogRequestRecord.URI, LogRequest masks query params matching these keys with trails.LogMaskVal:
 // - password
 //
-// if logger.Logger is nil, NoopAdapter returns and this middleware does nothing.
-func LogRequest(ls logger.Logger) Adapter {
-	if ls == nil {
+// If handler is nil, NoopAdapter returns and this middleware does nothing.
+func LogRequest(handler slog.Handler) Adapter {
+	if handler == nil {
 		return NoopAdapter
 	}
 
+	// NOTE(dlk): add constant values to include on every message here
+	ls := slog.New(handler.WithAttrs([]slog.Attr{
+		{Key: trails.LogKindKey, Value: trails.HTTPLogKind},
+	}))
+
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			uri := r.URL.Path
-			q := r.URL.Query()
-			if val := q.Get("password"); val != "" {
-				q.Set("password", "xxxxxxx")
-			}
+			start := time.Now()
 
-			query := q.Encode()
-			if query != "" {
-				uri += "?" + q.Encode()
-			}
+			writer := &requestLogger{ResponseWriter: w, status: http.StatusOK}
+			h.ServeHTTP(writer, r)
 
-			strs := []string{r.Method, uri}
-			val := r.Context().Value(IpAddrCtxKey)
-			if val != nil {
-				strs = append([]string{val.(string)}, strs...)
-			}
+			end := time.Since(start).Milliseconds()
 
-			ls.Info(strings.Join(strs, " "), nil)
-			h.ServeHTTP(w, r)
+			rec := newRecord(writer, r)
+			rec.Duration = end
+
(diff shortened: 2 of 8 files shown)


http/middleware/get_ip_address.go
```diff
@@ -10,8 +10,6 @@ import (
 	"github.com/xy-planning-network/trails"
 )
 
-const IpAddrCtxKey = "trails/middleware/ip-address" // TODO(dlk): change to key provided by app?
-
```
DavidLarsKetch, line 14: 🧹 An errant reference in `log_request_test.go` is removed in this PR, so clear out this too.

http/middleware/log_request.go
```diff
@@ -2,44 +2,167 @@ package middleware
...
+		IPAddr:         ip,
+		Method:         r.Method,
+		Path:           r.URL.Path,
+		Protocol:       r.Proto,
+		Referrer:       r.Header.Get(referrerHeader),
+		ReqContentLen:  contLen,
+		ReqContentType: r.Header.Get(contentTypeHeader),
+		Scheme:         r.URL.Scheme,
+		Status:         w.status,
+		URI:            uri.RequestURI(),
```
seedifferently, line 113: what's the difference between `r.URL.RequestURI()` and `r.RequestURI`, or are they the same?

http/middleware/log_request.go
```diff
@@ -2,44 +2,167 @@ package middleware
...
+	BodySize       int    `json:"responseBodySize"`
+	Duration       int64  `json:"duration"`
+	Host           string `json:"host"`
+	ID             string `json:"id"`
+	IPAddr         string `json:"remoteAddr"`
+	Method         string `json:"method"`
+	Path           string `json:"path"`
+	Protocol       string `json:"protocol"`
+	Referrer       string `json:"referrer"`
+	ReqContentLen  int    `json:"requestContentLength"`
```
seedifferently, line 78: IMO since we're omitting `request` on the names of the other request params (and it's implied) then these should just be `contentLength` and `contentType`

log.go
```diff
@@ -0,0 +1,19 @@
+package trails
+
+import "golang.org/x/exp/slog"
+
+const (
+	LogKindKey = "kind"
```
seedifferently, line 6: you prefer the germanic "kin" over the greek/latin "type" eh? 😄

http/middleware/log_request.go
```diff
@@ -2,44 +2,167 @@ package middleware
...
+	uri.RawQuery = q.Encode()
+
+	contLen, _ := strconv.Atoi(r.Header.Get(contentLenHeader))
+	id, _ := r.Context().Value(trails.RequestIDKey).(string)
+	ip, _ := r.Context().Value(trails.IpAddrKey).(string)
+
+	return LogRequestRecord{
+		BodySize:       w.bodysize,
+		Host:           r.Host,
+		ID:             id,
```
seedifferently, line 103: nice 👍

http/middleware/log_request.go
```diff
@@ -2,44 +2,167 @@ package middleware
...
+		slog.Int("status", r.Status),
+		slog.String("uri", r.URI),
+		slog.String("userAgent", r.UserAgent),
+	}
+}
+
+type requestLogger struct {
+	http.ResponseWriter
+	status   int
+	bodysize int
```
seedifferently, line 141: no lowerCamel on bodysize?

http/middleware/log_request_test.go
```diff
@@ -20,47 +24,109 @@ func TestLogRequest(t *testing.T) {
...
 			http.MethodPut,
-			"0.0.0.0",
+			ip,
 			&url.URL{Path: "/hitting/the/trails", RawQuery: "param=true"},
-			"0.0.0.0 PUT /hitting/the/trails?param=true",
+			newExpected(middleware.LogRequestRecord{
+				IPAddr: ip,
+				Method: http.MethodPut,
+				Path:   "/hitting/the/trails",
+				URI:    "/hitting/the/trails?param=true",
```
seedifferently, line 85: would be nice to see a test asserting `#fragment` it retained on `uri` as well

http/middleware/log_request.go
```diff
@@ -2,44 +2,167 @@ package middleware
...
+	Method         string `json:"method"`
+	Path           string `json:"path"`
+	Protocol       string `json:"protocol"`
+	Referrer       string `json:"referrer"`
+	ReqContentLen  int    `json:"requestContentLength"`
+	ReqContentType string `json:"requestContentType"`
+	Scheme         string `json:"scheme"`
+	Status         int    `json:"status"`
+	URI            string `json:"uri"`
+	UserAgent      string `json:"userAgent"`
```
seedifferently, line 83: I don't see `timestamp` anywhere -- is that something slog outputs automagically?

seedifferently (commented):
This is looking great! I expected this to be a lot more changes but instead it looks pretty clean to me 💯 

I gave a few comments/questions.

http/middleware/log_request.go
seedifferently: I had a chance to look at the source and it seems they are equivalent: https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/net/http/request.go;l=1052

Why wouldn't we favor `r.RequestURI`  here instead to avoid the extra hop into parse and hop back to unparse? What does processing through `net/url` gain us?

log.go
seedifferently: I meant this as a dumb joke, but now that I see `slog` has a `Kind` type I imagine that's where you might've been going with this 👍

http/middleware/log_request.go
seedifferently: Revisiting the `slog` docs, it looks like `time` is provided automagically? So if that's the case forget my request to rename it to `timestamp` -- no need to mess with its default (I've just always preferred the explicit `timestamp` when its a datetime so that there isn't the ❓ of whether there's ever both `time` and `date` fields).

http/middleware/log_request.go
seedifferently: Sorry, I'm only getting intermittent chances to look into this (which is probably annoying) -- but glancing at the 2nd half of that function linked above makes me wonder if the hop through `net/url` _is_ actually useful to appropriately account for uncommon request cases such as `CONNECT`: https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/net/http/request.go;l=1073-1080;bpv=0 

But maybe I just need some helping grokking what's going on here.

DavidLarsKetch: Yeah the little jig on lines 90-94 is the main reason given some values in the URI are masked.

log.go
DavidLarsKetch: Haha, nice. It's more that `type` is a reserved keyword so don't want any shenanigans running into it when naming fields, vars, etc.

http/middleware/log_request.go
DavidLarsKetch: It is a little strange they went with `time`, maybe it'll change before merging in to the std lib. I'll show in a forthcoming PR how to overwrite these default keys, so we can relitigate there if need be.

http/middleware/log_request.go
DavidLarsKetch: Not if there's a typo!

http/middleware/log_request.go
DavidLarsKetch: Glad you bring this up. What's your appetite on using `slog`'s grouping functionality? Cf., [`slog.Group`](https://pkg.go.dev/golang.org/x/exp/slog#Group), which has a plain text example, rewritten as JSON, those `req` fields would be an object:
```json
"req": { "method": "GET", "url": "localhost" }
```
Personally, I'd rather go this direction, but comes down to whether traversing these nested relationships is a PITA for CloudWatch or whether it's simply easier to keep it flat.

http/middleware/log_request_test.go
DavidLarsKetch: This sent me down a rabbit trail. Looks like fragments wouldn't be logged in any case? I haven't looked at the source code for how `*http.Request.URL` is formulated, specifically, but I'm seeing fragments are not included in part because the browser doesn't send it (Firefox; but also check out the docstring for [url.ParseRequestURI](https://pkg.go.dev/net/url#ParseRequestURI) which confirms that's the expected behavior).

Accordingly, what I use to construct URI ([url.URL.RequestURI](https://pkg.go.dev/net/url#URL.RequestURI)) doesn't include a fragment even if manually set in a `*url.URL`.

DavidLarsKetch:
@seedifferently one thing to call out is this PR (nor the forthcoming one) solves logging requests for assets. [Here is how we handle those requests](https://github.com/xy-planning-network/trails/blob/df24c7bbda46099134cbe790a2d871f9332a7457/http/router/router.go#L93-L103), which needs its own refactoring, in my opinion, especially if we want to use `embed` for any of it.

Given we aren't currently logging these requests, I've deprioritized that fix, but let me know if I should re-evaluate.

http/middleware/log_request.go
DavidLarsKetch: I went ahead with your suggestions, but it's worth a gander at `slog.Group`. It's a quick change if you like the way it formats logs.

http/middleware/log_request_test.go
seedifferently: Wow, great info. Thanks David

http/middleware/log_request.go
seedifferently: honestly/unfortunately it's going to be hard for me to know for sure until I get some stuff piped into cloudwatch to get a better sense of what the ergonomics are like. I can tell you right now that I prefer having to type out `contentType` a whole lot more than having to type out `requestContentType` or `{"req": {"contentType": ...}}`, but maybe once I get into cloudwatch I'll see that its magic 🪄 makes all of this super easy no matter how we've formatted it.

I say we go with what you have here now (flat & terse), and just extend me some grace to maybe want to take a 2nd pass at this once I get CW dashboards wired up 🤞 🙏

http/middleware/log_request.go
```diff
@@ -2,44 +2,167 @@ package middleware
...
+
+func (rl *requestLogger) Header() http.Header { return rl.ResponseWriter.Header() }
+
+// Unwrap exposes the underlying http.ResponseWriter.
+//
+// NOTE(dlk): include in order to simplify using [net/http.ResponseController].
+func (rl *requestLogger) Unwrap() http.ResponseWriter { return rl.ResponseWriter }
+func (rl *requestLogger) Write(b []byte) (int, error) {
+	size, err := rl.ResponseWriter.Write(b)
+	rl.bodySize += size
```
seedifferently, line 152: is this costly performance wise, with the only purpose being to get response body size? Response body size in logs is something that's rarely useful so I can live without it if this is a perf hit.

seedifferently:
> ... logging requests for assets

the hope is that CW logs are logging every request so that a full picture of the request journey can be painted when needed, but for now I agree it's not a high priority so can continue to punt on that

http/middleware/log_request.go
DavidLarsKetch: I don't have any benchmarking data to work with to make an assessment, but I can't think of a handler calling `Write` more than once (trails doesn't fwiw), which is what I naively conjecture is the way there'd be a meaningful perf impact. The most likely situation where that'd happen in our code base is if someone hands the `http.ResponseWriter` to a proc for writing directly to it. There's 1 case of that in `college-try`, not seeing one in `second-child` or `wall-to-wall`.

DavidLarsKetch:
Any further comments or LGTM?

seedifferently:
I'm happy with this @DavidLarsKetch! Thanks for cranking it out.

PR #40, opened 2023-04-21
Don't pass embedded fs.FS to http/template.TagPacker

DavidLarsKetch:


DavidLarsKetch:
@steven-klein this ended up being the fix. We can revisit embedding `client/dist` in the future, but not strictly necessary.

PR #41, opened 2023-04-26
Replace `log` with `slog`

DavidLarsKetch:
## Problem statement
We want to offload more of our logging duties to [`slog`](https://pkg.go.dev/golang.org/x/exp/slog). There are a few reasons to do this:
1. Our current logging is restrictive with the reliance on `LoggerContext`
2. Less maintenance
3. Becoming a part of the std lib, future teammates are likely to be familiar and comfortable with `slog`

We want the transition to be smooth, however, and not require a wholesale switch to `slog` straight out of the gate. This is because of the package's planned integration in the std lib as well as its newness to us.

## What this does
The topline to call out is that this PR requires no changes in our applications in order to utilize it 🎉 

### Updated
This PR switches out the internal logging library used in `logger` (which was `log`) with `slog`. It moves more responsibility for configuring `logger` to `ranger`.

Within that internal switch, `ranger` leverages 3 different [`slog.Handler`](https://pkg.go.dev/golang.org/x/exp/slog#Handler) implementations to toggle between depending on the application's configuration. These are:
- JSON encoded logs, when `LOG_JSON` is true or  `ENV` is prod or staging
- Plain text logs, when `LOG_JSON` is not true, `ENV` is not prod or staging, and the kind is HTTP router logs
- Tint text logs, when `LOG_JSON` is not true, `ENV` is not prod or staging, and the kind is not HTTP router logs (i.e., application and worker logs)

Lastly, the `logger.Logger` & `logger.SkipLogger` interfaces were separated for flexibility initially, but let's squash them together in an effort to continue a single path for using `trails`.

### Deleted 
- `logger/opt.go`:
    this functionality is no longer needed given the responsibility moves to `ranger` to setup a `*slog.Logger` first
- `logger.LogLevel`:
    use [`slog.Level`](https://pkg.go.dev/golang.org/x/exp/slog#Level) instead
- `logger.Logger.Fatal`:
    there's no `Fatal` for `slog.Logger` and there appear to be no uses in our repos, so let's remove it

### Fixed
Lastly, there's been an ongoing issue with logging file & line numbers, specifically for Faktory workers and Sentry. This PR recalibrates skipping by runtime frames for those cases. As well, it leverages the Sentry SDK for removing that package's own frames from the stack trace as well as those that the logger is already configured to skip. Notably, this means the Sentry package and trails will largely disappear from the top-level frames in a stacktrace, thereby surfacing the actual codepoint in the application that an error was executed. Compare [old](https://xy-planning-network.sentry.io/issues/4133328578/?project=5507546) vs. [new](https://xy-planning-network.sentry.io/issues/4133328578/?project=5507546); look specifically at the point where Sentry says the error occurred as well as the trace itself.

## Followups
- Checkout #42, which removes `WorkerLogger()`


http/middleware/log_request.go
```diff
@@ -24,16 +24,11 @@ const (
...
-	if handler == nil {
+func LogRequest(ls *slog.Logger) Adapter {
+	if ls == nil {
 		return NoopAdapter
 	}
 
-	// NOTE(dlk): add constant values to include on every message here
-	ls := slog.New(handler.WithAttrs([]slog.Attr{
-		{Key: trails.LogKindKey, Value: trails.HTTPLogKind},
-	}))
```
DavidLarsKetch, line 35: Moved to `ranger`

logger/logger.go
```diff
@@ -25,195 +29,125 @@ var (
...
+
+	l.l.Handler().Handle(context.TODO(), rec)
 }
 
-// Fatal writes a fatal log.
-func (l *TrailsLogger) Fatal(msg string, ctx *LogContext) {
-	if l.ll > LogLevelFatal {
-		return
+// ColorizeLevel adds color to the log level!
+func ColorizeLevel(groups []string, a slog.Attr) slog.Attr {
```
DavidLarsKetch, line 98: To minimize the changes in our fork of [tint](https://github.com/xy-planning-network/tint), I opted to implement a `ReplaceAttr` function to add to [`slog.HandlerOptions`](https://pkg.go.dev/golang.org/x/exp/slog#HandlerOptions). While that fork includes changing the color of the log message, this function here makes it so that fork doesn't include changes to changing the log level color.

profsmallpine (approved):
lgtm

PR #42, opened 2023-04-28
Adds BuildWorkerCore

DavidLarsKetch:
## Problem Statement
`ranger.New` sets up too much for our worker binaries, thus an application ends up duplicating some of the default setup found in `ranger`. A worker, specifically, does not need an HTTP router and all the accoutrement attending a router.

## What this does
This PR adds `ranger.BuildWorkerCore` which constructs a slim version of a `*Ranger` that does not have an HTTP router.

I opted for a whole `*Ranger` as the return value so that future changes are simpler; i.e., the function signature need not change. What led to this was how `college-try` uses the base URL constructed by the `*Ranger` in its Faktory workers while `second-child` does not. Trying to keep track of how each repo may be slightly different seemed like a bad idea.

### DIFF
--- a/http/resp/responder.go
+++ b/http/resp/responder.go
@@ -86,7 +86,8 @@ func NewResponder(opts ...ResponderOptFn) *Responder {
 		opt(d)
 	}
 
-	d.logger = d.logger.AddSkip(responderFrames)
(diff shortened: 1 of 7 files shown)


profsmallpine (approved):
lgtm

PR #45, opened 2023-05-11
Adds fix for sentry nil stacktrace and maintenance mode

profsmallpine:


logger/sentry.go
```diff
@@ -97,7 +97,7 @@ func (sl *SentryLogger) send(level sentry.Level, ctx *LogContext) {
 		}
 
 		if ctx.Data != nil {
-			scope.SetExtra("data", ctx.Data)
+			scope.SetContext("data", ctx.Data)
```
profsmallpine, line 100: SetExtra is deprecated: https://docs.sentry.io/platforms/go/enriching-events/context/#additional-data

ranger/config.go
```diff
@@ -22,6 +22,10 @@ type Config[U RangerUser] struct {
 	// FS is the filesystem to find templates in for rendering them.
 	FS fs.FS
 
+	// MaintMode determines how to configure ranger on ranger.New.
+	// If true, it skips setting up a database connection and routes to a maintenance page.
+	MaintMode bool
```
profsmallpine, line 27: Went with a config value so that the caller could determine the logic on when to enable

ranger/ranger.go
```diff
@@ -289,3 +314,19 @@ func (m Metadata) templateFunc() (string, func(key string) string) {
 }
 
 type ShutdownFn func(context.Context) error
+
+func maintModeHandler(p template.Parser, l logger.Logger, contact string) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		w.Header().Add("Retry-After", "300")
+
+		tmpl, err := p.Parse("tmpl/maintenance.tmpl")
```
profsmallpine, line 322: Skipped the responder as `resp.Html` is too heavy with it checking flashes and other aspects of an "app" response.

ranger/ranger.go
```diff
@@ -108,12 +105,40 @@ func New[U RangerUser](cfg Config[U]) (*Ranger, error) {
...
+	}
 	r.Router = defaultRouter(r.env, r.url, r.Responder, mws)
 	r.srv = defaultServer(r.ctx)
 
+	if cfg.MaintMode {
+		methods := []string{
+			http.MethodDelete,
+			http.MethodPatch,
+			http.MethodPost,
+			http.MethodPut,
```
seedifferently, line 130: why not cover all 9? https://pkg.go.dev/net/http#pkg-constants

DavidLarsKetch (commented):
Sentry updates look 👍 

I'm a bit torn on the maintenance mode setup. Functionality looks great (echo @seedifferently's question about methods). However, there's so few of the components in `Ranger` that mode needs and the additional toggling feels like we're straying from the "happy path" being left-aligned. Did you play around with just writing a new unexported func like `newMaintRanger` and returning that in `New`? Curious if you had any insights there.

profsmallpine:
@DavidLarsKetch good call. That was getting a bit too spaghetti. Little bit of duplicated code but this keeps it so that changes can be made without concern of breaking the all important ranger.New flow happy path.

profsmallpine:
I might need to sanity test this again, but when I go to `/cats` or w/e, I get redirected to `/`. I saw this in the docs and used it without much digging:

```
r := mux.NewRouter()
r.HandleFunc("/specific", specificHandler)
r.PathPrefix("/").Handler(catchAllHandler)
```

https://github.com/gorilla/mux#matching-routes

PR #47, opened 2023-05-28
Ensure maintenance mode handler responds properly with 503

seedifferently:
The maintenance mode handler was returning a `200 OK` response. This PR:

- Changes the response code to `503`
- Bumps the `retry-after` header to `600` (I figured that for our typical use-case of DB maintenance, `300` might be a bit low -- but I guess it's all pretty arbitrary)
- Adds a test

### DIFF
--- a/ranger/ranger_test.go
+++ b/ranger/ranger_test.go
@@ -1,0 +1,50 @@
+package ranger_test
+
+import (
+	"bytes"
+	"io/fs"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/xy-planning-network/trails/http/template"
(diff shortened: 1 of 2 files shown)


ranger/ranger_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/stretchr/testify/require"
+	"github.com/xy-planning-network/trails/http/template"
+	tt "github.com/xy-planning-network/trails/http/template/templatetest"
+	"github.com/xy-planning-network/trails/logger"
+	"github.com/xy-planning-network/trails/ranger"
+	"golang.org/x/exp/slog"
+)
+
+func TestMaintModeHandler(t *testing.T) {
```
profsmallpine, line 18: 🏅

profsmallpine (approved):
👍

DavidLarsKetch (approved):

PR #48, opened 2023-05-30
http/resp/responder: adds more logging for handleHtmlError

DavidLarsKetch:
Logs weren't initially included in order to put the responsibility on the caller of `*Responder.Html`, but that's proved not ergonomic and more likely to miss dispatching to Sentry.

### DIFF
--- a/http/resp/responder.go
+++ b/http/resp/responder.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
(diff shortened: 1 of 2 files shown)


profsmallpine (approved):
This looks good, I'm a little torn on the muddiness that returning an error that has been sufficiently handled adds to code here, but conclude having the error to do something seems necessary. All good here

PR #49, opened 2023-05-30
Configures gorm logger

profsmallpine:
I debated passing the config through ranger or something else so that it could be completely overrode, but opted to simply remove not found errors and remove color outside of dev. This creates the minimal amount of changes from what we have now while also not allowing more configuration we don't have an immediate use case for.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@
 [._]s[a-rt-v][a-z]
 [._]ss[a-gi-z]
 [._]sw[a-p]
+
+# Mac OS auto-generated files
+.DS_store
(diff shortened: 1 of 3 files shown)


DavidLarsKetch (approved):

.gitignore
```diff
@@ -18,3 +18,6 @@
 [._]s[a-rt-v][a-z]
 [._]ss[a-gi-z]
 [._]sw[a-p]
+
+# Mac OS auto-generated files
+.DS_store
```
seedifferently, line 23: note that `.gitignore` is typically case-sensitive, but this should for `.DS_Store` on macos due to its unique FS case sensitivity handling

PR #53, opened 2023-09-06
Go 1.21

DavidLarsKetch:
Go 1.21

Notably, replaces `golang.org/x/exp/slog` for `log/slog`

A later PR would introduce any changes the new [`maps`](https://pkg.go.dev/maps) and [`slices`](https://pkg.go.dev/slices) std lib pkgs enable.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,6 @@ jobs:
 
       - uses: actions/setup-go@v4
         with:
-          go-version: "1.21"
+          go-version-file: "go.mod"
 
       - run: go test ./... --cover -v -race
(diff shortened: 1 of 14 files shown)


.github/workflows/test.yml
```diff
@@ -15,6 +15,6 @@ jobs:
 
       - uses: actions/setup-go@v4
         with:
-          go-version: "1.20.4"
```
seedifferently, line 18: I'd recommend switching to the `go-version-file` ref like we've done here https://github.com/xy-planning-network/second-child/pull/575/files

seedifferently:
@DavidLarsKetch also, looks like Go 1.21.1 was just released 🙌  https://go.dev/doc/devel/release#go1.21.1

profsmallpine (approved):
🧹

steven-klein (approved):

PR #54, opened 2023-09-06
Logger path fix

DavidLarsKetch:
## Problem statement
Using the `exp` pkg for `slog` meant the API would change before landing in the std lib - and it has! As well, the pkg used for colorizing output made the jump from `exp` to std lib. The type switch to overwrite the source file & line number broke, accordingly.

## What this does
This PR replaces the type switch for an assertion given the std lib `slog` and `tint` pkg both coalesced on the same type.


seedifferently (approved):
awesome!

profsmallpine (approved):

steven-klein (approved):

PR #55, opened 2023-09-08
Adds DATABASE_MAX_IDLE_CXNS for configuring DB idle connection pooling

DavidLarsKetch:
## Problem statement
My research for the performance regression for [this Notion ticket](https://www.notion.so/xylabs/Address-ReconciliationProcedures-Diff-performance-regression-7bfb0299d89249a0b163b10276c30900?pvs=4) led to an interesting finding. The simplest, easiest "win" was to not close/discard idle DB connections. The handler returned 33% faster on average.

Currently, a query opens a database connection and closes it when it completes. There is no default connection pooling enabled.

## What this does
This PR adds an env var `DATABASE_MAX_IDLE_CXNS` which allows for keeping open a certain number of idle database connections. The number of actual connections can exceed the max, but when queries complete, the connection returns to the pool until the pool equals `DATABASE_MAX_IDLE_CXNS`. If the max has been reached, it closes the connection.

Without setting the env var, the default is `2` - the same default in `database/sql`.

NB: unless I'm mistaken, this kind of setting is not a PostgreSQL thing (which does have a concept of max connections, fwiw), and comes from `database/sql`. It is a feature of the connection pooling the std lib pkg provides.

I had a hard time finding where the `database/sql` default was overriden, but my guess is [`pgx`](https://github.com/jackc/pgx) since it does not use its pooling library by default.

## Discussion
The proc that was helped by this change runs a lot of small queries because of work performed across goroutines. Goroutines were the original performance solution The re-opening of those connections appears to have a demonstrable impact on the procs performance overall.

Refactoring the proc itself is an always an option so that it performs less queries. Those would be more complicated queries, so it appears simpler to me to not have this internal bottleneck.

I expect this change to have a blast radius larger than the one proc I was focused on (given this is run locally, I struggled to find another endpoint I could hit that was slow enough to also test against). I speculate this will have a positive impact on overall HTTP response times (e.g., `/benefits` is likely slow because of the number of SQL queries being performed, rather than there being 1 meaningfully slow query). Of course, the opposite could happen, so I'd like to set up a test to validate this hypothesis - something to discuss?

### DIFF
--- a/postgres/postgres.go
+++ b/postgres/postgres.go
@@ -19,14 +19,15 @@ const cxnStr = "host=%s port=%s dbname=%s user=%s password=%s sslmode=%s"
 
 // CxnConfig holds connection information used to connect to a PostgreSQL database.
 type CxnConfig struct {
-	IsTestDB bool
-	URL      string
-	Host     string
-	Port     string
-	Name     string
-	User     string
-	Password string
-	SSLMode  string
+	IsTestDB    bool
+	URL         string
+	Host        string
+	Port        string
+	Name        string
+	User        string
+	Password    string
+	SSLMode     string
+	MaxIdleCxns int
 }
 
 // Connect creates a database connection through GORM according to the connection config and runs all migrations.
@@ -43,7 +44,7 @@ func Connect(config *CxnConfig, migrations []Migration, env trails.Environment) 
 		c.Colorful = true
 	}
 
-	db, err := gorm.Open(postgres.Open(buildCxnStr(config)), &gorm.Config{
+	gormDB, err := gorm.Open(postgres.Open(buildCxnStr(config)), &gorm.Config{
 		Logger: logger.New(log.New(os.Stdout, "\r\n", log.LstdFlags), c),
 		NamingStrategy: schema.NamingStrategy{
 			NameReplacer: strings.NewReplacer("Table", ""),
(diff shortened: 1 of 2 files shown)


postgres/postgres.go
```diff
@@ -56,17 +57,24 @@ func Connect(config *CxnConfig, migrations []Migration, env trails.Environment)
 		return nil, err
 	}
 
+	db, err := gormDB.DB()
+	if err != nil {
+		return nil, err
+	}
+
+	db.SetMaxIdleConns(config.MaxIdleCxns)
```
DavidLarsKetch, line 65: I hate this GORM pattern, but yes, this is mutating the parent `gormDB` because of the pointers at hand.

DavidLarsKetch:
In particular, @seedifferently & @profsmallpine, curious if connection pools has come up with `second-child` or the recent infra moves.

profsmallpine (approved):
I forgot specifically why this came up, but during the archive migration to cdk/ec2, we were talking about changing the idle conns and punted into the future and the problem got resolved elsewhere. This lgtm

DavidLarsKetch:
@seedifferently Nothing jumps off the page w/r/t GORM about fussy behavior reconnecting to the DB when no connection is available in the pool. It's fairly lightweight, relying on `jackc/pgx`. If there's an issue in that library, well... I think we're heading into more tenuous territory so don't want to waste energy there.

seedifferently:
ok cool, thanks for taking a look @DavidLarsKetch -- sounds like what you've already got here is the best way forward

PR #56, opened 2023-09-13
inline brand related assets

steven-klein:
# What this does

- inlines a fallback png type favicon for Safari
- inlines a primary SVG favicon for the cool kids
- inlines the primary XYPN logo in templates
- remove manifest reference (don't have a manifest to serve - something we can do on a rainy day)
- remove theme-color meta (I'm not opposed to having this, it was just only in the maintenance and error templates and seemed odd.  It's another small touch we can revisit on a rainy day, like apple-touch-icon)


profsmallpine (approved):
inline ftw

ranger/tmpl/error.tmpl
```diff
@@ -36,7 +35,9 @@
     padding: .25rem .5rem;
   }
   #logo {
-    margin: 0 0 2rem 100px; 
+    margin: 2rem 120px; 
+    width: 150px;
+    height: auto;
   }
   #sash {
     background: #1F6DF4;
```
DavidLarsKetch, line 43: Since you're already in here, want to swap the old blue for the new `#2850BE`?

DavidLarsKetch (approved):
LGTM!

ranger/tmpl/error.tmpl
steven-klein: good catch!

PR #58, opened 2023-09-15
Add ID for sessions 

DavidLarsKetch:
## Problem statement
Two problems:
1. only authenticated sessions persist between requests; each unauthenticated request starts a new session
  - This is because `session.Session.Save` is not called for every session created, and instead was saved only when other methods like adding flashes or registering a user.

2. there's no unique identifier to track sessions across requests

## What this does
This PR adds a UUID to each session. And, it saves sessions for all requests during the `InjectSession` middleware.

### DIFF
--- a/http/middleware/log_request.go
+++ b/http/middleware/log_request.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/xy-planning-network/trails"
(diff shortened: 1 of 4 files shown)


profsmallpine (approved):
Really tightening things up today, noice

steven-klein (approved):

PR #63, opened 2023-09-18
Logs all requests to assets, client, and not found

DavidLarsKetch:
## Problem statement
Requests for assets, distributed client, and not found paths are not logged currently. This hinders observability and debugging.

## What this does
This PR includes the `middleware.LogRequest` middleware on the handlers for requests to filesystem resources - i.e., the built client and assets. There's a bit of refactoring in here to help with readability.

As well, requests ending up in `404` or `307` because of not found are logged.

### DIFF
--- a/http/router/router.go
+++ b/http/router/router.go
@@ -68,6 +68,7 @@ type Router interface {
 type DefaultRouter struct {
 	Env           string
 	everyReqStack []middleware.Adapter
+	logReq        middleware.Adapter
 	*mux.Router
 }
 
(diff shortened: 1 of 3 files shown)


profsmallpine (approved):

http/router/router.go
```diff
@@ -92,22 +93,28 @@ func (r *DefaultRouter) AuthedRoutes(
 // NewRouter constructs an implementation of [Router] using [DefaultRouter] for the given environment.
 //
 // TODO(dlk): use provided [fs.FS] and [http.FS] instead of [http.FileServer].
-func NewRouter(env string) Router {
+func New(env string, logReq middleware.Adapter) Router {
 	r := mux.NewRouter()
+	cacheControl := cacheControlMiddleware()
+
+	assetsServer := http.FileServer(http.Dir(assetsPublicPath))
```
steven-klein, line 100: I'm late to the party, and this is not a critical issue since we're currently avoiding static assets, but the current `assetsPublicPath` only exists in a local development environment where you have the full source. 

When Vite runs a build, static assets remain unprocessed by the build pipeline, but are copied from `client/public/${asset_name}` into `client/dist/${asset_name}`.  This makes local testing for these paths tricky, because even when you've run a local client build `npm run build -- --mode=staging` your local server is still serving them from `client/public/${asset_name}` which will not exist in a deployed docker image.

In other words, when the client is in a development mode via `npm run dev`, the server can find static assets in `assetsPublicPath`, but in all other cases where the client has been built those assets will be found in `clientDistPath`.

PR #64, opened 2023-09-18
http/router: refactor

DavidLarsKetch:
- Removes `Router` interface; renames `DefaultRouter` to `Router`: reason, YAGNI
- Improves code readability by unnesting functions in favor of assigning to an identifier

### DIFF
--- a/http/router/router.go
+++ b/http/router/router.go
@@ -23,77 +23,18 @@ type Route struct {
 	Middlewares []middleware.Adapter
 }
 
-// A Router handles many [Route], directing HTTP requests to the appropriate endpoint.
-type Router interface {
-	// AuthedRoutes registers the set of Routes as those requiring authentication.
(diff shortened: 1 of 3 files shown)


profsmallpine (commented):
Pondering this, seems like the router is something that is likely to be replaced so rather than this existing to swap out by the caller, it provides an interface to satisfy or at least makes explicit what we are using from the mux router. If the golang routing proposal goes through and we decide to use that, I see having this wrapped as an interface a win.

DavidLarsKetch:
@profsmallpine I can see the value of keeping the interface if there's a transitional step to a new package that means upgrading your `trails` version doesn't mean changing the implementation of routing. In my opinion, if we're stripping out `mux` at some point, I'd be looking to re-do the API anyway since the methods are tightly coupled to `mux`. Of course, that can be incremental, too. So, I could see bringing back an interface when necessary. But, not a strong opinion one way or another.

PR #65, opened 2023-09-20
exports postgres.MigrateUp, call in *ranger.Ranger.Guide

DavidLarsKetch:
## Problem statement
Migrations run prematurely in app startup. In all use cases, app initialization means constructing a `*ranger.Ranger` before emitting properties from it for other app services, procedures, etc. Because migrations run when establishing the database connection, this is too early in an app's startup. Migrations should be the last operation before starting an app due to the unknowns reversing their schema and data changes.

As well, there are no longer any uses for running migrations when building a worker, so this change of use case needs to be updated in `ranger`.

[Notion ticket](https://www.notion.so/xylabs/Spike-DB-migrations-f73c55740a1546459f6d28d40a162da4?pvs=4)

## What this does
This PR exports `postgres.MigrateUp` so a `*ranger.Range` can call it in `Guide`. `Guide` is the final step in app startup since it begins accepting HTTP requests, so provides the best opportunity for delaying running migrations.

This PR removes `ranger.WorkerConfig` given the use for running migrations is no longer necessary.

### DIFF
--- a/postgres/migrate.go
+++ b/postgres/migrate.go
@@ -35,7 +35,7 @@ func (m Migration) execute(db *gorm.DB) error {
 	return nil
 }
 
-func migrateUp(db *gorm.DB, migrations []Migration) error {
+func MigrateUp(db *gorm.DB, migrations []Migration) error {
 	// Ensure public schema exists
 	ensurePublicSchema(db)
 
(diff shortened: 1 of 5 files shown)


profsmallpine (approved):
Nice!

seedifferently:
🎉

PR #66, opened 2023-09-22
asset handling for static and bundled assets

steven-klein:
# What this does

- consolidates existing client side asset HTTP handlers into a single handler serving from `client/dist/`
- replaces TagPacker with a similar function that is focused only on generating the path to any client asset and updates the templates accordingly

## Why

We recently realized that we have a problem serving client side static assets.  This is in part due to the fact that those assets are at different paths when in different environments and/or client side development modes (build vs serve).  Finding the balance between having a backend that is moderately opinionated on where assets should live and setting a client side bundler's config to fit can get confusing quickly.  

This PR aims to make a few things more consistent, namely, when you are doing local development, it's the responsibility of the dev server (`vite serve`) to get the client assets to the browser - which is good for working with reactive libraries like Vue where having hot reloading greatly benefits the DX.  Additionally, it moves some of the path complexity that exists between environments and development modes away from needing to be configured or handled -- see template changes on setting the asset path based on the environment.

## Notes

- While digging into this, it became clear this won't be as simple as starting with a basic Vite config and working from there.  Vite is widely adopted, but also heavily opinionated, we don't fit any of the primary use cases for Vite.  We actually fall into what they consider an `appType` of "custom" - because we do three things:
  1. bring our own backend
  2. have a multipage application design
  3. arrange our client side entrypoint file structure in a way that does not mirror the routes.  i.e. they all live in `src/pages`
- It's probably a good thing to have Trails be opinionated about where the client assets directory is expected to be and not feel the need to make it configurable.  We're more likely to swap Vite out one day, than adopt so kind of Golang backend framework and every client side asset toolkit is configurable for a reason - few setups are alike and to be widely adopted you still need to be flexible.
- I was still able to reduce the complexity of the [vite.config.ts](https://github.com/xy-planning-network/college-try/commit/c4fee1b90ef6f91403a221fa4f64dfc719c737e4) by quite a lot.  Though most of these changes aren't all that relevant.  The important changes are:
  1. keep the `base` path consistent to `client/dist` (it's unintuitive, but it keeps Trails and Vites base paths the same when they each serve assets)
  2. remove custom output handling - Trails should be able to care a little less about where exactly an asset ends up when built as long as it's inside the `client/dist` directory.

## TODO

> If this gets the general go ahead

- [x] cleanup dead code
- [x] add tests for AssetURI
- [x] optional - make the origin host for assets configurable


http/router/router.go
steven-klein, line 1: I considered whether there might be a need for a static asset specific handler - primarily for a different cache control configuration, but I don't believe I've ever needed that in a project.

http/template/tag_packer.go
```diff
@@ -60,3 +63,41 @@ func TagPacker(env trails.Environment, filesys fs.FS) func(string, bool) html.HT
 		}
 	}
 }
+
+// AssetURI encloses the environment and filesystem so when called executing a template,
+// emits valid URI for client side static and bundled assets.
+//
+// TODO:
+//
+// - configurable OriginHost would allow for support of a CDN and custom Vite server origin
```
steven-klein, line 72: The goal here was to move this into the direction of supporting a configurable Host for assets.  If we do that, then technically we shouldn't even need the `env.IsDevelopment()` case, as the host env would just be set to `http://localhost:8080/` in our local environments.  Having a configurable host means CDN support as well as customized origin with a dev server like `vite serve` - we don't need that with Tax going away, but it would have been nice to have when Tax existed.

http/template/tag_packer.go
```diff
@@ -60,3 +63,41 @@ func TagPacker(env trails.Environment, filesys fs.FS) func(string, bool) html.HT
...
+			// match hashed files bundled by Vite
+			filename := strings.TrimSuffix(assetPath, filepath.Ext(assetPath))
+			fileExt := filepath.Ext(assetPath)
+
+			// Note: where assetPath = assets/GetDashboard.js
+			// glob = client/dist/assets/GetDashboard-*.js
+			glob := fmt.Sprintf("%s/%s-*%s", assetsBase, filename, fileExt)
+			matches, err := fs.Glob(filesys, glob)
+
+			if errors.Is(err, path.ErrBadPattern) || len(matches) == 0 {
```
steven-klein, line 96: This one is tricky, but I feel like if we don't match the pattern when looking for a hashed asset, we should leave the path unchanged.  This lets the logs show the request path if it 404's.  I say if, because a static asset will not match this glob pattern, since they are not hashed and should resolve as expected.

The alternative to the glob strategy is to consume a manifest, which has some pros/cons, but to leverage it should probably be geared more towards the "optimization" it provides by getting all of an entry points dependencies on to the page with `<script module="preload">`.

ranger/tmpl/layout/vue.tmpl
```diff
@@ -7,7 +7,11 @@
     var InitialVueProps = {{ .Data.props }}
   </script>
 
-  {{ packTag .Data.entry false }}
+  {{ if isDevelopment }}
+  <script src="{{ asset (print "src/pages/" .Data.entry ".ts") }}" type="module"></script>
+  {{ else }}
+  <script src="{{ asset (print "assets/" .Data.entry ".js") }}" type="module"></script>
+  {{ end }}
```
steven-klein, line 14: I accept that this might be an unexpected change, but once I landed on it, I felt like there was improved clarity on the problem that is, the entry point exists at a different path when the development mode is changed, and it avoids needing to "configure" the sub directory via Trails or Vite.  They both just need to be configured to serve their respective assets under `client/dist`.

It also avoids hiding away special handling for file extensions. i.e. `.ts` vs `.js`, and while we don't do it, a similar case for could exist for `.scss` and `.css`.  

Finally, FWIW, once I started using a tool chain that came with a dev server (browserify, webpack dev server, vite serve), I've always ended up keeping this logic in my template.

ranger/tmpl/layout/authenticated_base.tmpl
```diff
@@ -14,7 +14,7 @@
     {{ template "additionalHeadScripts" }}
 
     {{ if not isDevelopment }}
-    {{ packTag "app" true }}
+    <link rel="stylesheet" href="{{asset "assets/style.css"}}">
```
profsmallpine, line 17: Nitpick: `<link rel="stylesheet" href="{{ asset "assets/style.css" }}">` for readability on the spacing around `{{}}`

ranger/tmpl/layout/vue.tmpl
profsmallpine: Waffling on whether this should get hidden in trails to keep the implementation simpler in the tmpl, but I think I agree that it puts the problem/solution in plain view which is probably the right call.

profsmallpine (commented):
This seems like a sane approach, and I like having the assetURI attempt to be less smart than tag packer and just do one thing. Are we able to remove the tag packer function and change the filename?

ranger/tmpl/layout/vue.tmpl
steven-klein: Yeah, I tried it a few ways., but this one out in part because we aren't putting a lot of asset related tags into the templates anyway.  It also helps decouple from the client a bit more, giving a Trails user a little more freedom to structure their client src as preferred.  With a configurable origin, it goes further and opens up the flexibility to use Trails with a client tool chain other than Vite since the only thing Vite specific becomes the pattern match for hashed assets.

http/template/tag_packer.go
seedifferently: wrt globbing, it makes me wonder:
1. How brittle is this? I'm concerned about footguns. i.e. What are the odds of matching the wrong file because somebody didn't name something _just right_?
2. Probably not an issue in our world, but I'd be curious to know what kind of perf hit all this globbing causes. Maybe some memoization is worth exploring here?
3. I'd be interested to hear a little more about your thoughts on utilizing a manifest.

http/router/router.go
seedifferently: What's our current avenue for cache busting assets? Does the hash change on every build or is it more involved?

seedifferently (commented):
Considering our unique setup of `not really SPA` and `not really SSR` this doesn't seem too bad to me.

I had a couple of Qs that'd help me get some clarity on some things.

http/template/tag_packer.go
steven-klein: @seedifferently 

1.  It's reasonably resilient based on current experience, but has the potential to become problematic.  Trails matches against the single main stylesheet and the JS entrypoint files for every page load.  Since we keep all entrypoint files in a single src directory, you can't create a naming collision, but because Vite/Rollup will create chunks for composable and component files as well, one of those could cause a collision.   ex: We have numerous `Main.vue` files that are likely to be chunked to `Main-xxxxx.js`, if you created an entrypoint file amed `Main.ts` - you'll have a hard time getting the right file to match.
2. It definitely could be a perf issue at some point, but I'm not terrible concerned right now. We're only making 2-3 match attempts: `style-xxxxxx.css`, `GetEntrypoint-xxxxxx.js` and potentially `favicon.ico`.  Consuming a manifest at bootstrap and storing it in memory for an indexed lookup would certainly resolve any issues there.
3. I think the glob match is nice in the sense that it doesn't rely on the toolchain creating a manifest.  Vite does it out of the box with a single config option, but that wasn't true with Webpack the last time I used it, and I'm uncertain what other alternatives like Parcel.js are doing.  That being said, I've assumed we'd either need or want the benefits eventually, once we've decided they outweigh the added complexity.  One way I've leveraged the manifest in the past is to use its presence in the file system as an indicator of the mode my client side is in.  Since a client side dev server mode is not necessarily the same as being in ENV=development, finding the asset paths based on the existence of a `manifest.json` can at time be very helpful.

The major advantages to consuming a manifest:

- separates client side mode from server side ENV
- faster and more resilient asset matching for entrypoints
- we can drop the conditional <script> in the template, as the manifest key is typically the `source` path
- (optional): The manifest provides a list of chunks that are dependencies of your entrypoint, so we can revist something like TagPacker to spit out all the client side assets we know the page is going to request asynchronously and tag them with the preload attribute.  This would include stylesheets as chunks as well, but based on our setup I'm not certain there's much of a win there.

Cache busting:

I have not reconfirmed this, but the standard for years with assets that go through the build pipeline has been to hash the chunked files contents.  This keeps you from busting the cache on files that don't change often, like vendor dependencies in node_modules, while also ensuring that chunks can be optimized as needed on each build.

A static asset would require a filename change to ensure a busted cache, since it is not hashed and is served by the same handler.

seedifferently: ok great thanks for filling in the gaps for me. I'm 👍 with this. My only other bit of feedback might be to log a warning if `len(matches) > 1` so there's some visibility around that scenario should it ever happen.

ranger/default.go
```diff
@@ -106,7 +107,8 @@ const (
 )
 
 var (
-	defaultBaseURL = "http://" + DefaultHost + DefaultPort
+	defaultAssetsURL = "/"
```
steven-klein, line 110: I stuck with the existing default assets URL which is the root path.  As setting it to the defaultBaseURL with a scheme and host will cause CORS issues for `connect.xyplanningnetwork.com`.  While not all CDN's can be configured to include a CORS header, the one I think we're most likely to use, Cloufront, does as does the Vite dev server.

ranger/tmpl/error.tmpl
```diff
@@ -5,7 +5,7 @@
...
+    <link rel="icon" sizes="48x48" href="{{ asset " favicon.ico" }}" />
```
steven-klein, line 8: During QA I discovered that Safari still doesn't want to load this, so I'm just falling back to the project supported `favicon.ico` for now.  Safari v17 was released recently and has SVG favicon support in technical preview, so hopefully it will be resolved, and we can remove this entirely.

steven-klein:
@profsmallpine @seedifferently @DavidLarsKetch Origin support has been added via the `ASSETS_URL` env var.  Given that we don't need to use it in prod/staging at the moment and setting a value with a scheme and host would cause issues for FAA this implementation means it does need to be set for local dev to ASSETS_URL=http://localhost:8080.   

You can see the Vite supporting setup here: https://github.com/xy-planning-network/college-try/pull/861

http/template/assets.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	assetsBase = "client/dist"
+)
+
+// AssetURI encloses the environment and filesystem so when called executing a template,
+// emits valid URI for client side static and bundled assets.
+//
+// The asset origin defaults to "/" when it is not configured and is always set with a trailing slash.
+func AssetURI(origin *url.URL, env trails.Environment, filesys fs.FS, l logger.Logger) func(string) string {
```
DavidLarsKetch, line 25: Late to the party... but!

Does it make more sense to return an error alongside the path instead? The closure signature would be `func(string) (string, error)`, accordingly.

Then, if we're wanting to have some observability on this specific issue either:
1. error out and fall back to the error template in `ranger` where `resp.Html` will log the error, or,
2. throw another template function together that logs the error and throws it away, passing along the `string` the `AssetURI` closure returns.

I don't want to make this more complicated than it's worth with a purist critique. But, to me, including an initialized service in order to have observability functionality is not simple. It nudges the function towards more complexity along the lines of "AssertURI, toggles around the env to put together a path to an asset in the filesys that incorporates the origin, and logs when globbing fails to return a single asset".

DavidLarsKetch: I don't think we have any template functions that return errors, so to a quick example: https://go.dev/play/p/3fptfpnruDd

steven-klein: I do think there's a pretty decent argument for returning an error and having it fallback to the error template, particularly since finding multiple matches is likely going to mean a broken page that has no UI to speak of.  If we move in the direction of consuming the manifest, that could increase the surface area of new types of errors that occur, so having the ability to return errors of any kind an know that the error template is rendered when things go wrong sounds like a good idea.

steven-klein: Screenshot 2023-09-27 at 4 56 26 PM

This feel very handy.

http/template/assets.go
```diff
@@ -0,0 +1,69 @@
...
+			// glob = client/dist/assets/GetDashboard-*.js
+			glob := fmt.Sprintf("%s/%s-*%s", assetsBase, filename, fileExt)
+			matches, err := fs.Glob(filesys, glob)
+
+			if errors.Is(err, path.ErrBadPattern) {
+				return "", fmt.Errorf("%w: for asset path %s", path.ErrBadPattern, assetPath)
+			}
+
+			if len(matches) > 1 {
+				return "", fmt.Errorf("%w: for asset path %s", ErrMatchedAssets, assetPath)
```
seedifferently, line 57: 💯

seedifferently (approved):
looks like a nice improvement to me!

DavidLarsKetch (approved):
Awesome LGTM!

PR #73, opened 2024-02-28
[ PROPOSAL ] Generic DB calls

DavidLarsKetch:
## Proposal

This PR rewrites the `DatabaseService` interface by providing two wrappers around a `*gorm.DB`: `Simple[T]` and `Robust[T]`. In both cases, `T` identifies the return value of terminal methods like `First`, `Find`, `Update` and so on. `Simple[T]` will be made available to HTTP handlers, which generally do not (i.e., ought not) mutate data and simply only read it. `Robust[T]` will be made available to procs so they can read, write, update and delete.

### Problem statement
There are two goals:
1. leverage generics to simplify initializing values returned from queries; and,
2. restrict the flexibility of GORM's API so we more closely approach "one way" to access the database.

#### flexibility
Let's start with (2). We care about restricting the flexibility of GORM's API for two main reasons. First, it curtails using some facets of the type loose-ness of the GORM API. Second, it closes off the need to enforce patterns in PR review. Right now, we use our shared experience and that's becoming a pain point as more of us work in `college-try`. Currently, our handlers use `DatabaseService` while procs use `*gorm.DB`. Let's talk about the issues there.

`DatabaseService` is a solution to (2) by limiting what kinds of queries can happen and what can be expressed by those queries. I find the solution to be too divorced from both SQL syntax and GORM's API that I generally always copy/paste from an example in order to ensure I'm using a function signature correctly. It doesn't seem necessary, in any case, to have another _way_ of expressing a query, as the functions, and instead can solve the problem another way.

On the procs side, there is no existing solution to the flexibility of GORM's API and you will find between `college-try` and `second-child` meaningfully different ways of querying. That can entail the strings passed into many of GORM's methods or even which methods are called (e.g., `Pluck` vs. `Select`).

The goal, then, is to replace _both_ of these with `Simple[T]` subbing in for `DatabaseService` and `Robust[T]` subbing in for `*gorm.DB`.

#### generics
(1) is a new goal and is a QoL improvement made possible

### What this does
This PR is an outline of `Simple[T]` and `Robust[T]`. In a few methods, I've scratched out how those methods are implemented so the idea is clearer.

`First`, `Find`, `Count`, `Paged`, `Update`, `Updates`, `Create`, `Delete`, `Exec`, `Raw` finish a query. When called, they return a `T` with the results of the previously built query. If they fail, they gracefully wrap errors in a `trails` defined error.

`First` returns a single item; `Find` returns a list. For both methods, `T` would be just the single type rather than a slice of `T`. That is:
```go
user, err := NewSimple[domain.User]Where("id = ?", id).First()

users, err := NewSimple[domain.User].Where("access_state = ?", domain.AccessGranted).Find()
``` 

Other methods build a query. These leverage the same builder pattern in GORM. Generally speaking, these query building methods will simply call the GORM equivalent. Only where we want to add in some functionality would they diverge from GORM.

#### TODOs
- [ ] The specifics of `Begin` and `Commit` given terminal methods don't return a `Robust[T]`.
- [ ] Confirm how to implement the builder pattern.
- [ ] Confirm whether `Model` and `Table` conflict

### DIFF
--- a/postgres/service.go
+++ b/postgres/service.go
@@ -1,27 +1,151 @@
 package postgres
 
 import (
+	"database/sql"
+	"errors"
+	"fmt"
 	"math"
 
+	"github.com/xy-planning-network/trails"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
-// DatabaseService sets up the interface to be used at the handler/middelware level. These should be straightforward
-// calls that allow us to skip creating a procedure method for the most basic database interactions. At the procedural
-// layer, the *gorm.DB struct is available directly for more complex composition. This has the intended functionality
-// that we are not testing the database in handlers, while it is tested directly at the procedural layer.
-type DatabaseService interface {
-	CountByQuery(model any, query map[string]any) (int64, error)
-	FetchByQuery(models any, query string, params []any) error
-	FindByID(model any, ID any) error
-	FindByQuery(model any, query map[string]any) error
-	PagedByQuery(models any, query string, params []any, order string, page int, perPage int, preloads ...string) (PagedData, error)
-	PagedByQueryFromSession(models any, session *gorm.DB, page int, perPage int) (PagedData, error)
+type Simple[T any] struct {
+	db *gorm.DB
 }
 
-// PagedData is returned from the PagedByQuery method. It contains paged database records and pagination metadata.
-type PagedData struct {
-	Items      any   `json:"items"`
+func NewSimple[T any](db *gorm.DB) *Simple[T] {
+	return &Simple[T]{db}
+}
+
+func (db *Simple[T]) Debug() *Simple[T] { return db }
+
+func (db *Simple[T]) Distinct(args ...interface{}) *Simple[T] { return db }
+
+func (db *Simple[T]) Find() (T, error) { return nil }
+
+func (db *Simple[T]) First() (T, error) {
+	var dest T
+	err := db.db.Model(&dest).First(&dest).Error
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return fmt.Errorf("%w", trails.ErrNotFound)
+	}
+
+	if err != nil {
+		return fmt.Errorf("%w: %s", trails.ErrUnexpected, err)
+	}
+
+	return nil
+}
+
+func (db *Simple[T]) Group(name string) *Simple[T] { return db }
+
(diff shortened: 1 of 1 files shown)


seedifferently:
This is really cool! Definitely seems like it enhances the ergonomics/dev QoL for working with gorm.

My primitive knee-jerk reaction however is that I'm wary of abstractions so if I'm going to go back and rework a bunch of my queries I'd rather move in the opposite direction -- moving toward _less_ abstraction on top of my database libs rather than more.

I'm also not convinced I see your two problem statements as actual problems...at least problems I ought to care that much about. e.g. if the goal is to `restrict the flexibility of GORM's API so we more closely approach "one way" to access the database` why not make it a documented convention instead of a coded abstraction?

But from a readability and ergonomics perspective I'm very much 😍

DavidLarsKetch:
> documented convention instead of a coded abstraction

I'll start by saying that we'll still need conventions with this proposal, so it isn't a panacea. Hopefully, this minimizes the conventions. That's mainly because of the `string`s and variadic parameters GORM accepts. Off the top of my head, I know some places multiple conditions are placed in a single `Where` and some places that's spread across multiple `Where` calls:
```go
db.Where("access_state = ? OR date_joined IS NOT NULL AND created_at > ?", AccessState, "2 weeks ago")
db.Where("access_state = ?, AccessState).Or("date_joined IS NOT NULL").Where("created_at > ?", "2 weeks ago")
```

The gist of the value of coded abstractions for me in this case is type safety. With the changes to GORM's signatures proposed, there isn't the option to misalign with the patterns we've reached consensus on.

To me there's a meaningful difference between articulating conventions about a programming language, on the one hand, and a library, on the other. The need for the former is just part and parcel of programming. The need for the latter, for me, means you're just working with a poorly designed library (which we are) and can be mitigated with thoughtful interventions (which I vainly believe this PR is 😉 ). It's a spectrum; if this were Rails I was trying to wrap behind an abstraction - that might be a bad endeavor. But, this is GORM; it's a bad API.

>  moving toward less abstraction on top of my database libs rather than more

I agree! Less abstraction would be better, I just don't see it as an option. Maybe the problem is better stated as:

"We've accepted GORM as our database lib and it doesn't work for us without meaningful hiccups and discrepancies across codebases."

We already have a partial solution to this with `DatabaseService`, so I find this PR is a refactor and extension of the path we've set out on. Removing this in our HTTP handlers is of course an option, but now we've reintroduced another point where conventions must enforced (e.g., "don't start a transaction in a handler").

PR #76, opened 2024-04-11
Runs 'go get gorm.io/driver/postgres'

profsmallpine:


seedifferently (approved):
👍 maybe also a `go mod tidy` to clean up the old refs?

PR #79, opened 2024-06-14
adds http/req

DavidLarsKetch:
## Problem statement
There's more than 1 way to parse, on the one hand, JSON payloads and, on the other, query params in HTTP requests. A project must import `github.com/gorilla/schema` and `github.com/go-playground/validator/v10`, configure the respective packages and then handle the errors propagating from those. There should only be one way to do this and no discrepancies about configuration. There should be - that is - a uniform way of parsing and validating HTTP request payloads.

We require developers to have an in-depth knowledge about the std lib and 3rd-party packages enabling us to parse and validate, but the flexibility has proved to be neither necessary nor desirable. That is, it introduces more decisions to be made rather than less.

Of notable concern are errors. Errors hit us when developing and in production code. The information we receive ought to be different in each case. But, because of the ephemeral nature of some of these errors (i.e., you only run across them every once in a while), our current patterns only up the ante with regards to how much you need to know about `json`, `schema`, or `validator`.

## What this does
This PR provides a new package to abstract parsing and validating HTTP request payloads: `http/req`. That package provides a struct that can, in turn, parse and validate JSON data or query param data. In both cases, the diversity of errors each package may return are conformed to a standard type: `ValidationErrors`.

To get to a `ValidationError`, or set of them, `http/req` inspects the types of errors `schema` and `validator` may return and coerces them to the package's own type. This puts the message an error reports into a standard format so that a JSON request or a query param request return similarly worded and formatted messages, in lieu of being bound to the implementation details of one package or another (i.e., `json` vs. `schema` vs. `validator` saying the same thing in different ways).

`ranger` takes `http/req` one step further by providing helpers that actually write HTTP statuses based upon parsing or validation errors. The intended use case is with API handlers (web handlers may have their their own helper or helpers, but scoping this PR to API handlers felt clear for a first pancake). What's particularly helpful about this is marshaling `ValidationErrors` to JSON when in dev mode. The goal is is a clean, pretty-printed format that enables developers to take direct action on.

As a last tidbit, this PR configures `schema` to ignore unknown keys in order to mirror `json`'s default configuration. These are not configurable on a codebase, but could be made available in a future PR.

### Codebases
With this PR, `college-try` and `second-child` delete parsing HTTP requests ([here](https://github.com/xy-planning-network/college-try/blob/master/http/api/parse.go) and [here](https://github.com/xy-planning-network/college-try/blob/master/http/web/parse.go) and [here](https://github.com/xy-planning-network/second-child/blob/master/http/api/parse.go)) and special validation configuration ([here](https://github.com/xy-planning-network/college-try/tree/master/services/validator)). `ranger.Ranger.ReqParser`, `ranger.Ranger.ParseBody` and `ranger.Ranger.ParseQueryParams` provide all the needed functionality.

### API responses
There's been conversation around a constant, top-level schema for API responses amongst our applications. This PR does not resolve this conversation and I would like to close off any discussion here. Instead, I find this as a constituent part of that broader goal that actually makes the bigger one possible. It's simple enough to say there's an `errors` key present for every non-2xx, but it's not simple to say why our API may return `400` or what messages look like when validation requirements have not been met. This PR attempts to solve these latter concerns.

### Enumerables
This PR moves `college-try`'s enumerables into `trails`. This covers the [domain definition](https://github.com/xy-planning-network/college-try/blob/master/domain/enum.go) and [validator service](https://github.com/xy-planning-network/college-try/blob/398dae20cabbdfeeab89f5b960317ad780d793f9/services/validator/validator.go#L128-L153), which would be deleted


http/req/doc.go
```diff
@@ -0,0 +1,16 @@
+/*
+Package req provides ergonomics for handling an HTTP request.
```
seedifferently, line 2: is `req` really doing any other "handling" than parsing? maybe `http/parse` is more explicit?

http/req/doc.go
```diff
@@ -0,0 +1,16 @@
...
+
+Package req provides a helper for parsing payloads in an HTTP request.
+It supports JSON-encoded payloads and payloads encoded in query parameters.
+In both cases, package req expects to parse payloads into a pointer to a struct.
+That struct ought to leverage the apporpriate struct tags for performing two tasks.
+First, matching keys in the payload to fields on the struct.
+Second, for validating the payload's data meets requirements.
+
+By leveraging req, handlers can get data out of an HTTP request into its application specific structs.
```
seedifferently, line 11: imo would be nice to have example parser usage code in the doc (or wherever makes the most sense)

seedifferently (commented):
Cool -- at a cursory glance this seems like a nice solution for our current parsing 🤹. 

Was there any evaluation of other packages in this space aside from validator/v10 and gorilla/schema? (Not suggesting we use something else, just curious about the state of play for solving this sort of problem in the Go community, and if other approaches were eval'd and why you settled on these.)

Left a few comments/questions below.

(note: opting for the `Comment` review option here because I'd like to see feedback from others before this is ✅'d and 🚀'd)

http/req/doc.go
DavidLarsKetch: Yeah great suggestion, I was thinking of going with [Go's examples](https://go.dev/blog/examples). Too simple?

ranger/ranger.go
```diff
@@ -136,8 +140,111 @@ func (r *Ranger) Context() (context.Context, context.CancelFunc) { return r.ctx,
...
+			ranger.Error(err.Error(), &logger.LogContext{Error: err, Request: r})
+		}
+
+		return false
+	}
+
+	var verrs *req.ValidationErrors
+	if errors.As(err, &verrs) {
+		args := []resp.Fn{resp.Code(http.StatusUnprocessableEntity)}
+		if !ranger.env.IsProduction() {
```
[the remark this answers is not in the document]
DavidLarsKetch: I personally wouldn't like for validation errors to be returned in production, yet. The intended consumer of this data is only the devs, presumably, looking at the JSON payload in the browser console. I can anticipate a feature in the future that builds off of this and translates these packaged-up error messages into a UI for team members, but I'm hoping to scope this PR to the former.

ranger/ranger.go
```diff
@@ -136,8 +140,111 @@ func (r *Ranger) Context() (context.Context, context.CancelFunc) { return r.ctx,
...
+// ParseBody expects r.Body to be JSON-encoded data.
+// If r.Body is incorrectly formatted JSON, ParseBody writes 400.
+//
+// ParseBody expects the decoded data to pass "validate" struct tags defined on payload's type.
+// If values in the data do not meet validation requirements, ParseBody writes 422.
+//
+// In all cases, if an unhandled error occurs, ParseBody writes 500.
+//
+// If control over response codes is desired, use ranger.ReqParser.ParseBody.
+func (ranger *Ranger) ParseBody(w http.ResponseWriter, r *http.Request, payload any) bool {
```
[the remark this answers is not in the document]
DavidLarsKetch: I can see that, I left the method name broad since I do see this being the same signature for parsing/validating HTML forms (toggle around `Content-Type`). We've got just a handful of those on portal, so I wasn't too worried about including them here, but can walk that out if it seems helpful now.

ranger/ranger.go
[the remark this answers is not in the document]
seedifferently: cool makes sense, if that's the case though I'd suggest the more explicit `if ranger.env.IsDevelopment` then over the current "is not prod" check.

http/req/doc.go
seedifferently: that's actually what I had in mind too

http/req/doc.go
DavidLarsKetch: It's validating the data parsed as well, but nothing beyond that yet. Data from an HTTP request being the primary entity managed by this package, `req` felt broad enough to include those 2 functions, but I'm not married to it.

DavidLarsKetch:
> Was there any evaluation of other packages in this space aside from validator/v10 and gorilla/schema?

Clarifying that question, is your question about whether packages other than these 2 were evaluated?

If it is, then, I didn't look into replacing either since I find moving this parse/validation functionality over to `trails` is the first step towards such an evaluation. Swapping out either of those would entail lots of changes (from a pure line count perspective) and so could be a larger project than we want to take on right now.

In my ideal world, those packages (and whether there even are any dependencies) become implementation details. It might be down the road we replace setting `json`, `schema` and/or `validator` struct tags with our own (e.g., `req:"myField,required"` defines the casing in JSON- and URL-encoding and identifies it can't be `null`) to enable simpler refactoring.

Let me know if that answers the question!

seedifferently:
> I didn't look into replacing either since I find moving this parse/validation functionality over to trails is the first step towards such an evaluation.... In my ideal world, those packages (and whether there even are any dependencies) become implementation details.

Makes sense, thanks!

ranger/ranger.go
```diff
@@ -136,8 +140,111 @@ func (r *Ranger) Context() (context.Context, context.CancelFunc) { return r.ctx,
...
+		}
+
+		if err := ranger.Json(w, r, args...); err != nil {
+			ranger.Error(err.Error(), &logger.LogContext{Error: err, Request: r})
+		}
+
+		return false
+	}
+
+	if err := ranger.Json(w, r, resp.Err(err), resp.Code(http.StatusInternalServerError)); err != nil {
```
profsmallpine, line 190: This is interesting from a logging perspective. Currently, we have a lot of:

```
if err := h.parseRequest(r, req); err != nil {
		h.Json(w, r, Err(err), Code(http.StatusBadRequest))
		return
	}
```

which logs the error and send it to sentry. If any of the scenarios fail (decoding, validating), we get an error. This has often been the preferred behavior since we control the shape of the payload so if an error is triggered in either decoding or validating, these have been high value notifications since I usually failed an edgecase on the frontend.

Since these errors do not get passed back to the calling code, we can't pass the responsibility to the calling code with the current implementation. In archive's case, an error is logged to sentry in every single case of failed parsing/validate flow. Additionally, if decoding succeeded, data is passed to sentry as well. I would be happy to default to logging the error from archive's perspective, but the issue of forwarding data remains as I don't want the entire payload passed to sentry if it has anything that shouldn't show up in logs/sentry. Portal has a handful of scenarios where nothing is logged so maybe those should log or maybe this needs more configurability. It's nice if there is any issue when it shows up as a red log on development as well.

http/req/validate.go
```diff
@@ -0,0 +1,110 @@
...
+
+	return validator{v}
+}
+
+// validate checks the fields on structPtr match the rules set by "validate" struct tags.
+// On success, validate returns no error.
+// On failure, validate translates each issue to a ValidationError,
+// returning them all as ValidationErrors.
+func (v validator) validate(structPtr any) error {
+	err := v.valid.Struct(structPtr)
```
profsmallpine, line 45: Name-spacing with `valid` adds confusion here. I think that embedding directly `validator.Struct(structPtr)` or `v.Struct(structPtr)` removes the unhelpful `.valid` step call as that appears more like an attribute.

ranger/ranger.go
```diff
@@ -136,8 +140,111 @@ func (r *Ranger) Context() (context.Context, context.CancelFunc) { return r.ctx,
...
+	if errors.Is(err, trails.ErrBadFormat) {
+		if err := ranger.Json(w, r, resp.Code(http.StatusBadRequest)); err != nil {
+			ranger.Error(err.Error(), &logger.LogContext{Error: err, Request: r})
+		}
+
+		return false
+	}
+
+	var verrs *req.ValidationErrors
+	if errors.As(err, &verrs) {
```
profsmallpine, line 177: Since the error that is passed back from `.validate` is not a pointer, this conditional will not get triggered from validate issues.

profsmallpine (changes requested):
Not sure on the best approach for logging and logged data when decoding succeeds but validation fails. I've got some ideas but not sure on the goals from portal's perspective. Happy to chat more if you want to see this get out of WIP before I'm out

profsmallpine:
> In my ideal world, those packages (and whether there even are any dependencies) become implementation details. It might be down the road we replace setting `json`, `schema` and/or `validator` struct tags with our own (e.g., `req:"myField,required"` defines the casing in JSON- and URL-encoding and identifies it can't be `null`) to enable simpler refactoring.

Fwiw, I do see that as the path forward. If we want to turn validation errors into useful client side errors, it would probably be easier to roll our own convention rather than trying to turn this validator's package into client messages. If we don't find ourselves wanting that, supporting a small subset of validation rules ourselves seems too simple to justify relying on a 3rd party.

ranger/ranger.go
DavidLarsKetch: I'm not sure I totally follow. If you'd like to log the error in some cases, `ranger.ReqParser.ParseBody` provides direct access to the library so the response can be controlled as well. Is it that you'd like a `ranger` method handling the response writing to also log the error?

For Portal's use case, it looks like the dominant pattern is `h.Json(w, r, Code(http.StatusBadRequest))`. I don't really see a strong throughline on `college-try` for those that diverge from the pattern (i.e., those that do `h.Json(w, r, Err(err), Code(http.StatusBadRequest))` and seems like they should all be refactored out. I haven't seen a need to raise these errors thus far, but we've also had many different clients historically, which haven't been as rock solid as our own ;)

It sounds like, though, configurability is important between the two projects, or, at least, for dev environments. What about a env var that disables or enables logging so there's some flexibility?

profsmallpine:
After sitting with this today, I do not find the juice worth the squeeze. I like the simplicity of keeping the parsing and validating close at hand rather than pushing it into trails. I experimented with getting the same functionality (different error codes for different failure types) by simply updating the parse request in each project. This was more straightforward as I wasn't having to error check to figure out which type of failure scenario was occurring, further separating what is happening from the error handling. Having a little bit of copy/paste is preferred here until we know what exactly what problem we'd like to solve by having a single source of truth for this code. I'm not swayed by the setup of gorilla and validator in each package as it's not something I've thought about for many years given the `.parseRequest` has become the 1 way to parse/validate.

Moving what is happening into trails hasn't been the win we had once envisioned given the pain of debugging. https://github.com/xy-planning-network/college-try/pull/1305 is a good example where trails changes the request under the hood, and our custom encoding was not possible with a trails response. I'd rather walk through the code more linearly when debugging something like this.

DavidLarsKetch:
> This was more straightforward as I wasn't having to error check to figure out which type of failure scenario was occurring

Can you elaborate a bit more, I'm not sure I follow the what the win was here?

> Having a little bit of copy/paste is preferred here until we know what exactly what problem we'd like to solve by having a single source of truth for this code.

Sure, the primary issue is: the need for more precise error codes differentiating parsing (`400`) versus validation (`422`) and this need meaning more boilerplate. By moving it into `trails`, we can provide a helper that does all of it for us.

In the discussion we're having, 2 other issues have emerged.
1. Wonder whether the expansive API surface of `validator` is helpful for us (including the overlapping functionality between `schema` and `validator` on `required` tags)
1.  Divergence around `h.Json(w, r, Err(err), Data(data))` and whether there should be a uniform way our APIs respond to requests

> Moving what is happening into trails hasn't been the win we had once envisioned given the pain of debugging.

Sorry to hear it. Let's roll stuff back if it's no longer helpful.

profsmallpine:
The win is the simplicity of starting with the "dumb" approach, defaulting to copy/paste as opposed to trying to unify the code into trails which is more complex. Parse goes from:

```
func (h *Handler) parseRequest(r *http.Request, reqStructPtr any) error {
	if err := json.NewDecoder(r.Body).Decode(reqStructPtr); err != nil {
		return err
	}

	if err := validator.New().Struct(reqStructPtr); err != nil {
		return err
	}

	return nil
}
```

to (as psuedocode as you could use resp.Fns here instead of WriteHeader and handle logging)

```
func (h *Handler) parseRequest(w http.ResponseWriter, r *http.Request, reqStructPtr any) error {
	if err := json.NewDecoder(r.Body).Decode(reqStructPtr); err != nil {
		w.WriteHeader(http.StatusBadRequest)
		return err
	}

	if err := validator.New().Struct(reqStructPtr); err != nil {
		w.WriteHeader(http.StatusUnprocessableEntity)
		return err
	}

	return nil
}
```

Getting the error messages in something more uniform isn't a problem I want to solve given the how non-user-friendly they are from the validator specifically and not something I would be translating on the client.

DavidLarsKetch:
I think I'm selling myself short in trying to summarize the problem, so apologies if I'm being unclear. I want to amend that.

Nevertheless, the thrust of what you've responded to is the additions in `ranger`. That's easy enough to remove, of course, if we prefer copy-pasting your code snippet. To build off yours - and bracketing the question of logging errors - it'd look something like:
```go
func (h *Handler) parseRequest(w http.ResponseWriter, r *http.Request, reqStructPtr any) error {
	err := h.ReqParser.ParseBody(r.Body, reqStructPtr)
	if errors.Is(err, trails.ErrBadFormat) {
		h.Json(w, r, Code(http.StatusBadRequest))
		return err
	}
	
	if errors.Is(err, trails.ErrNotValid) {
	        h.Json(w, r, Code(http.StatusUnprocessableEntity))
		return err
	}

	if err != nil {
	       h.Json(w, r, Code(http.StatusInternalServerError))
	}
	
	return nil
}
```

What I'm selling myself short on is how `http/req` translates away from errors the `json`, `validator` and `schema` packages to a uniform format (`ValidationError`). `ranger` provides those errors as JSON when in dev or staging envs alongside mapping the sentinel errors to HTTP status codes. I take this error translation and regurgitation as a DX problem for the sake of this PR. That is, I find it helpful to have clarity, brevity and consistency between the errors those packages present when I'm building a new feature on the client and I've mis-stepped on naming things, typing things, and so on.

This work can aspire to formatting errors into a usable shape for end users (i.e., for `/staff` endpoints). I wouldn't expect `trails` to drive the UI, instead would expect `trees` to do that. But, I would expect `trees` to work with a simple interface instead of the diversity of errors and their messages our API currently leverages. How that gets accomplish is an open question, of course, and don't intend to resolve it here. For example, you're correct to call out `validator` doesn't provide nice messages. That's easy to fix should the groundwork here be merged in. Then, when we know what we're shooting for, what's useful for our stakeholders, maybe the change happens in `trails` or on `trees` to turn the default messages into something more useful.

profsmallpine:
I love the elegance that this provides, and the normalized error response is really tidy. However, I don't find it all that helpful practically. The logs as they come through now sufficiently if you had something like this:

```
func (h *Handler) parseRequest(w http.ResponseWriter, r *http.Request, reqStructPtr any) error {
	if err := json.NewDecoder(r.Body).Decode(reqStructPtr); err != nil {
		if h.Env().IsDevelopment() {
			h.Logger.Error(err.Error(), &logger.LogContext{Request: r})
		}

		h.Json(w, r, resp.Code(http.StatusBadRequest))
		return err
	}

	if err := h.Services.Validator.Struct(reqStructPtr); err != nil {
		if h.Env().IsDevelopment() {
			h.Logger.Error(err.Error(), &logger.LogContext{Request: r})
		}

		h.Json(w, r, resp.Code(http.StatusUnprocessableEntity))
		return err
	}

	h.Json(w, r, resp.Code(http.StatusInternalServerError))
	return nil
}
```

e.g.:

Screenshot 2024-06-18 at 6 54 26 PM

This is all I actually need to know that I passed `query` as a param that doesn't exist in my spelled out payload. The same is true for validation errors without any special wrapping. I haven't come across instances where logging the existing error message is not sufficient to quickly fix my mistake.

When it comes to the client side errors side of things, the only problem I know I want to solve is for a string to land in our concept of a flash message in rare occasions. If we were trying to tie error messages to input fields so that the email error message is directly under the email input for example, then this approach with the validator package would have a little bit of wings.

However, if this chunk of code isn't doing what I want, I have to drop into `.ParseBody()` and then I have to start grokking the error conditionals of something like `if errors.Is(err, trails.ErrBadFormat) {`. Then I have to drop into what a `ranger.req` is and then `p.Validate`, all the time looking for special error types that trigger certain conditionals. This is a lot of steps down the rabbit hole. It reminds me of looking at a rails controller, and once you find yourself literally 16 levels deep, you loose the thread of what happens based on what is returned. I'd rather be able to dot right into the validation package (better yet, we do it ourselves) so that I'm not so many steps removed from the code that is being run here.

PR #82, opened 2024-06-25
http/req: flatten

DavidLarsKetch:
Based on @profsmallpine's feedback, this PR provides an alternative factoring of the code. In short, error handling happens in `http/req.ParseBody` and `http/req.ParseRequestParams` so this functionality is closer to the surface of the package's public methods. This removes one layer of drilling down into files in `trails` whenever parsing & validating behavior doesn't match with expectations.

### DIFF
--- a/http/req/decode.go
+++ b/http/req/decode.go
@@ -9,35 +9,18 @@ import (
 	"github.com/xy-planning-network/trails"
 )
 
-type queryParamDecoder struct {
-	decoder *schema.Decoder
-}
-
-func newQueryParamDecoder() queryParamDecoder {
+func newQueryParamDecoder() *schema.Decoder {
 	dec := schema.NewDecoder()
 	dec.IgnoreUnknownKeys(true)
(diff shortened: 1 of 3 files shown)


DavidLarsKetch:
Testing some more, I typo'd a `validate` struct tag and discovered `validator` panics. Cool. That's now handled in https://github.com/xy-planning-network/trails/pull/79/commits/1acf3e4419405f97e676603d3fb6b2999470612c. I'll figure out how this PR would incorporate the change if it seems like we're headed in the right direction.

In any case, I don't endorse this PR. But, I recognize the issue of nesting will be a blind spot for me as package author, so defer to everyone else. Handling those panics suggests we'll be caught out at times by our dependencies and the quality of their code. Keeping that concern below the surface and limiting its exposure to devs is desirable, in my opinion. I don't believe we all share the same level of familiarity with and comfort using `schema`, `validator`, and even `json`.

PR #85, opened 2024-08-20
sentry v0.11.0 -> v0.28.1; resolves http/resp.GenericErr skip frame bug

DavidLarsKetch:
Bump that sentry version up.

There's a small bug with `http/resp.GenericErr` that it captures the responder pkg's frame for `GenericErr` instead of the frame before it, which would be the caller's frame. This PR resolves that.


DavidLarsKetch:
I think I overestimated what Sentry was going to do with error unwrapping, but it doesn't appear to change anything about our usage. So, at least we're now current on our sdk version.

logger/sentry.go
```diff
@@ -25,9 +25,10 @@ type SentryLogger struct {
 // routing messages to the DSN provided.
 func NewSentryLogger(env trails.Environment, l Logger, dsn string) Logger {
 	err := sentry.Init(sentry.ClientOptions{
-		Dsn:          dsn,
-		Environment:  env.String(),
-		IgnoreErrors: []string{"write: broken pipe"},
+		Dsn:           dsn,
+		Environment:   env.String(),
+		IgnoreErrors:  []string{"write: broken pipe"},
+		MaxErrorDepth: 2,
```
DavidLarsKetch, line 31: It doesn't appear we wrap more than twice in our projects, so this seems like a sane default in line with Sentry's advice to keep it brief.

I also didn't see a change in exception handling when I wrapped a bunch of errors and changed this number, so 🤷

PR #88, opened 2024-10-18
Attempt to puplate tags from log data when sending to Sentry

seedifferently:
In Archive, our `data` items such as `source_id` would be _significantly_ more useful as tags in Sentry than as data.

Rather than adding a separate `SentryTags` field to the `LogContext` struct, I wondered if we could get away with some lazy conversion.

### DIFF
--- a/logger/sentry.go
+++ b/logger/sentry.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strconv"
 	"time"
 
 	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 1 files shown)


logger/sentry.go
```diff
@@ -96,6 +97,15 @@ func (sl *SentryLogger) send(level sentry.Level, ctx *LogContext) {
 			scope.SetContext("data", ctx.Data)
 		}
 
+		// attempt to populate tags from Data
+		tags := convertMapAnyToString(ctx.Data)
+		for k, v := range tags {
+			// Sentry tag keys/values are limited to 32 and 200 chars respectively
+			if len(k) <= 32 && len(v) <= 200 {
+				scope.SetTag(k, v)
```
profsmallpine, line 105: What about something like to pass along whatever we can?

```suggestion
	maxK := 32
	maxV := 200

	if len(k) < maxK {
		maxK = len(k)
	}

	if len(v) < maxV {
		maxV = len(v)
	}

	scope.SetTag(k[:maxK], v[:maxV])
```

seedifferently: cool yeah works for me

PR #92, opened 2024-11-21
logger: only prints *http.Request.Body or Form if in dev

DavidLarsKetch:
This PR makes `logger.Logger` environment aware by adding a new param when constructing its `logger.TrailsLogger` implementation. That `Environment` is factored in `logger.LogContext`. Now, when not in dev, the context's request's body and form do not print.

It is useful once in a blue moon to have the raw `*http.Request.Form` or `*http.Request.Body` in the server's logs (as opposed to watching requests in browser dev tools), so opted to keep this around instead of removing entirely.

Fixes DEV-3053

### DIFF
--- a/http/resp/responder_opt_test.go
+++ b/http/resp/responder_opt_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


seedifferently:
🫡

PR #95, opened 2024-12-03
unexports appPropsKey, exposes interacting with it through AppProps

DavidLarsKetch:
This PR enables resolving DEV-3094 according to the std lib's approved pattern. [context/Context.Value](https://pkg.go.dev/context#Context.Value) suggests using, what I'm interpreting as, a getter/setter pattern and not exposing the actual const used to store values in a `context.Context`.

The extra guarding, typing, etc. may appear overkill for our uses, but, there's an argument to be made that if we had followed this advice in the first place, we wouldn't have had the unexpected effect we do have of overwriting values in a `context.Context`. So, taking our lumps on this one.

If we're liking this pattern in general, the next step is to move the other keys over to a similar approach. I'd look into the possibility of generics here, but not likely it'll cut down on boilerplate since the core decision to be made with these getters/setters is how to merge together/overwrite an existing value set in a `context.Context` with a new value.

Throwing the actual `map[string]any` into a new type will give us future flexibility, in my opinion, for attaching methods to the type should they become necessary.

### DIFF
--- a/http/resp/response.go
+++ b/http/resp/response.go
@@ -393,7 +393,7 @@ func Vue(entry string) Fn {
 		}
 
 		props := map[string]any{"initialProps": init}
-		if val := r.r.Context().Value(trails.AppPropsKey); val != nil {
+		if val := trails.AppPropsFromContext(r.r.Context()); len(val) > 0 {
 			props["appProps"] = val
 		}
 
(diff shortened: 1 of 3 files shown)


DavidLarsKetch:
Checkout https://github.com/xy-planning-network/college-try/pull/1609 for what this means for our apps

steven-klein (approved):
LGTM!   

I would be interested in having a setter method like `trails.AppProps.Set(key, value)` for ergonomics, but isn't something that seems necessary with our limited usage of AppProps currently.

PR #97, opened 2024-12-16
Fix HTTP Referer header name

seedifferently:
I was trying to figure out why I wasn't ever seeing any referrer info in our logs, then realized we have better spellcheck than the original author of the spec. (see https://en.wikipedia.org/wiki/HTTP_referer)


DavidLarsKetch (approved):
Classic `Referer`.

PR #101, opened 2025-03-25
[DRAFT] Add `resp.Funcs` to pass in additional template functions on response

DavidLarsKetch:
Sometimes, you just need some functions defined at response time to be available to HTML templates being rendered. But, there's no way to do so. This also enables a bit of "function overloading" by replacing a function defined already in the responder's parser.


profsmallpine (approved):
I like this direction, being able to add parser funcs is nice to have in general ✅

PR #104, opened 2025-04-15
[DEV-3407] enables setting PostgreSQL schema for DB

DavidLarsKetch:
Moves away from hardcoding `public` in different calls setting up the DB. We can find a more elegant API down the line, but this gets the job done!

For an example: https://github.com/xy-planning-network/college-try/pull/1921


profsmallpine (approved):
Sick!

PR #105, opened 2025-04-25
[DEV-3502] Proposal: `postgres.DB` wraps GORM

DavidLarsKetch:
Here's the API:
```go
type DB struct {
	// Has unexported fields.
}

func Connect(cfg Config) (*DB, error)

// Terminal methods
func (db *DB) Count() (int64, error)
func (db *DB) Create(value any) error
func (db *DB) Delete(value any) error
func (db *DB) Exists() (bool, error)
func (db *DB) Find(dest any) error
func (db *DB) First(dest any) error
func (db *DB) Paged(page, perPage int) (PagedData, error)
func (db *DB) Update(values map[string]any) error

// Clauses
func (db *DB) Distinct(args ...any) *DB
func (db *DB) Group(name string) *DB
func (db *DB) Joins(query string, args ...any) *DB
func (db *DB) Limit(limit int) *DB
func (db *DB) Offset(offset int) *DB
func (db *DB) Or(query string, args ...any) *DB
func (db *DB) Order(order string) *DB
func (db *DB) Select(columns ...string) *DB
func (db *DB) Scope(scope Scope) *DB
func (db *DB) Table(name string) *DB
func (db *DB) Unscoped() *DB
func (db *DB) Where(query string, args ...any) *DB

// Transactions
func (db *DB) Begin(opts ...*sql.TxOptions) *DB
func (db *DB) Commit() error
func (db *DB) Rollback() error

// helpers
func (db *DB) Debug() *DB

// Raw SQL access
func (db *DB) Exec(sql string, values ...any) error
func (db *DB) Raw(dest any, sql string, values ...any) error

// GORM-ness
func (db *DB) Model(model any) *DB
func (db *DB) Preload(query string, args ...any) *DB
```

With regards to querying, this removes these methods from `*gorm.DB` (there's more that aren't relevant to our purposes):
```go
func (db *DB) Clauses(conds ...clause.Expression) (tx *DB)
func (db *DB) CreateInBatches(value interface{}, batchSize int) (tx *DB)
func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, batch int) error) *DB
func (db *DB) FirstOrCreate(dest interface{}, conds ...interface{}) (tx *DB)
func (db *DB) FirstOrInit(dest interface{}, conds ...interface{}) (tx *DB)
func (db *DB) Having(query interface{}, args ...interface{}) (tx *DB)
func (db *DB) InnerJoins(query string, args ...interface{}) (tx *DB)
func (db *DB) Last(dest interface{}, conds ...interface{}) (tx *DB)
func (db *DB) Not(query interface{}, args ...interface{}) (tx *DB)
func (db *DB) Omit(columns ...string) (tx *DB)
func (db *DB) Pluck(column string, dest interface{}) (tx *DB)
func (db *DB) RollbackTo(name string) *DB
func (db *DB) Row() *sql.Row
func (db *DB) Rows() (*sql.Rows, error)
func (db *DB) Save(value interface{}) (tx *DB)
func (db *DB) Scan(dest interface{}) (tx *DB)
func (db *DB) ScanRows(rows *sql.Rows, dest interface{}) error
func (db *DB) Session(config *Session) *DB
func (db *DB) Take(dest interface{}, conds ...interface{}) (tx *DB)
func (db *DB) Transaction(fc func(tx *DB) error, opts ...*sql.TxOptions) (err error)
func (db *DB) Update(column string, value interface{}) (tx *DB)
func (db *DB) UpdateColumn(column string, value interface{}) (tx *DB)
func (db *DB) UpdateColumns(values interface{}) (tx *DB)
```

Removing them fell into one of 3 camps:
1. YAGNI
    - We haven't needed it, and neither will our future selves
    - Highlighting `Having` as it does have 1 very nice use case in Portal. I'm personally a fan of this clause and can expect more as Portal Admin features mature. That being said, it is avoidable.
2. One way to do things
    - For example, `Pluck` and `Select` are nearly equivalent and `Select` covers all cases that `Pluck` does (and more). So remove the less useful one.
    - Or another example: `Updates`, `Update`, `UpdateColumn` and `UpdatedColumns`.
        - With the advent of partial updates in `college-try`, updating a single is becoming increasingly rare. Writing your update in a map is slightly more to type, but removing these
            1. simplifies decisions about methods to call and
            2. is less rewriting when an update goes from 1 column to 2.
3. Not SQL enough
    - GORM is not just an ORM, but also a query 
    - For example, `Omit` has no equivalent in SQL

### DIFF
--- a/postgres/db.go
+++ b/postgres/db.go
@@ -317,17 +317,26 @@ func (db *DB) Paged(page, perPage int64) (pd PagedData, err error) {
 		return PagedData{}, fmt.Errorf("%w: %s", trails.ErrUnexpected, err)
 	}
 
-	zero := big.NewFloat(0)
-	totalRecFl := new(big.Float).SetInt(big.NewInt(totalRecords))
+	// NOTE(dlk): use math/big for accurate float64 division.
+	totalPages := new(big.Float).SetInt(big.NewInt(totalRecords))
 	perPageFl := new(big.Float).SetInt(big.NewInt(perPage))
 
-	fl := big.NewFloat(0).SetMode(big.AwayFromZero)
-	if totalRecFl.Cmp(zero) != 0 && perPageFl.Cmp(zero) != 0 {
-		fl = fl.Quo(totalRecFl, perPageFl)
+	// NOTE(dlk): guard divison by zero.
+	zero := big.NewFloat(0)
+	if totalPages.Cmp(zero) != 0 && perPageFl.Cmp(zero) != 0 {
+		totalPages.Quo(totalPages, perPageFl)
+	}
+
+	// NOTE(dlk): We want rounding up, but Int64 rounds towards zero
+	// and RoundingMode doesn't change this.
+	// So, add one when it truncates incorrectly to get rounding up to the ceiling.
+	var acc big.Accuracy
+	pd.TotalPages, acc = totalPages.Int64()
+	if acc == big.Below {
+		pd.TotalPages += 1
 	}
 
 	pd.TotalItems = totalRecords
-	pd.TotalPages, _ = fl.Int64()
 
 	return pd, nil
 }
--- a/postgres/db_test.go
+++ b/postgres/db_test.go
@@ -988,8 +988,6 @@ func (suite *DBTestSuite) TestPaged() {
 	suite.Require().Nil(err)
 	suite.Require().Equal(int64(len(groups)), actual.TotalItems)
 	suite.Require().Equal(int64(11), actual.TotalPages)
-
-	// Arrange
 }
 
 func (suite *DBTestSuite) TestPreload() {


postgres/db.go
```diff
@@ -0,0 +1,701 @@
...
+
+	if err != nil && errSQLSyntax.MatchString(err.Error()) {
+		return fmt.Errorf("%w: %s", trails.ErrNotValid, err)
+	}
+
+	if err != nil {
+		return fmt.Errorf("%w: %s", trails.ErrUnexpected, err)
+	}
+
+	if res.RowsAffected == 0 {
```
profsmallpine: This one surprised me. Typically, `.First` is where we are worried about something not being found and where gorm returns an error. But `.Find` can return zero results and this is not an error state. Is this intentional? I see the extra handling in portal `if err != nil && !errors.Is(err, trails.ErrNotFound) {` so assuming it is intentional. This is going to be hard for me to get used to given that it different from current functionality. What's the strategy here that makes it worth changing?

profsmallpine: Doing some digging, it appears that `len(data) == 0` being something to handle is more the exception than the rule and it actually changes the logic in some places where the extra handling is now applied when it probably shouldn't be. e.g. groups#ExtendMeetings

DavidLarsKetch: Right, there'll be more `!errors.Is(err, trails.ErrNotFound)`. The reasoning is this two-fold: (1) eliminates a lot of `if len(data) == 0` code blocks; and (2) makes `First` and `Find` have identical behavior instead of one having this additional check and the other not.

Definitely call out any cases where it looks

DavidLarsKetch: Copying from chatting with @steven-klein:

Every Find ought to be followed up with at least `if err != nil`, so you're halfway there to ignore `ErrNotFound` if you so choose. Same guidance applies to `Update` and `Exec`. I understand other ORMs that GORM copies from don't have the error-first approach that Go does and so it's a bit foreign.

I'll grant the empty result set isn't an error in a strict sense. The query ran and nothing returned. No issues there. But, having `postgres.DB` wrap up the state in an error enables simpler logic on caller's side given our conventions of handling errors.

To reduce it down, the gist of my stance is: "hey it isn't annoying you have to check for errors and `if len(data) != 0`?!? What if you could do that in one line?"

profsmallpine: I don't really see this as an argument around go being an error first approach, but that most ORMs only error when the API semantics say “exactly one row must exist”. This is the case with `.First()` but not `.Find()` or `.Delete` etc, and I imagine devs coming from sqlalchemy, django orm, Sequelize, activerecord, etc.. would be caught out by this choice. If we want this type of path, I would update the API to have `UpdateOne()`, `DeleteOne()`, etc.. Let's let it percolate over the weekend and we can make a decision monday morning on whether to keep this or revert to the existing behaviour.

steven-klein (changes requested):
I bumped into an issue with `Paged` and the `totalPages` count while doing QA on the Portal migration that looks like it needs to be resolved here.

PR #117, opened 2026-03-19
[DEV-4368] go 1.25.1 -> go 1.26.0; go fix

DavidLarsKetch:
Also, ran `go fix` to modernize usage.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/xy-planning-network/trails
 
-go 1.26.0
+go 1.26.1
 
 require (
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/xy-planning-network/trails
 
-go 1.25.1
+go 1.26.0
```
seedifferently: I think you want either `go 1.26` or `go 1.26.1` here (the former will automatically reference the latest patch version)

seedifferently (changes requested):
one small thing otherwise lgtm

